PR #6, opened 2015-12-01
Fix deallocator for malloc-ed memory

jnabhan:
The previous version sets the address to zero before freeing it so memory allocated with unsafe is never actually released.

PS: We can also just set address to 0 after freeing the memory. If doing it in this current order actually has a purpose, then we can keep the extra variable addressToFree.


peter-lawrey:
A good catch, thank you.

PR #9, opened 2016-02-10
Code quality fix - Classes and methods that rely on the default system encoding should not be used. 

faisal-hameed:
This pull request is focused on resolving occurrences of Sonar rule 
squid:S1943 - Classes and methods that rely on the default system encoding should not be used. 
You can find more information about the issue here: 
https://dev.eclipse.org/sonar/rules/show/squid:S1943
 
 
Please let me know if you have any questions.
 
 
Faisal Hameed

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/BytesInternal.java
+++ b/src/main/java/net/openhft/chronicle/bytes/BytesInternal.java
@@ -35,13 +35,13 @@ import java.io.OutputStream;
 import java.io.UTFDataFormatException;
 import java.nio.BufferOverflowException;
 import java.nio.BufferUnderflowException;
-import java.nio.charset.StandardCharsets;
 import java.text.SimpleDateFormat;
 import java.util.Date;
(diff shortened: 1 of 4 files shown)


leventov:
It complicates source code without need, in a line like "NaN".getBytes() it's obvious that only latin alphabet letters are used, that have same codes in all reasonable encodings (I think)

faisal-hameed:
@peter-lawrey Which Charset you would prefer UTF_8 or ISO_8859_1 ?

peter-lawrey:
ISO_8859_1  makes it clear you don't expect more than one byte per
character.

faisal-hameed:
@peter-lawrey I have updated this PR with charset ISO_8859_1.

peter-lawrey:
This change has been merged manually. All the projects have been update to reflect this change (not just Bytes)

faisal-hameed:
@peter-lawrey Ok thanks.

PR #23, opened 2016-09-08
Synchronize MappedBytes.acquireNextByteStore().

scottkidder:
This fixes the major race condition with multiple appenders. Please review for performance.

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/MappedBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/MappedBytes.java
@@ -137,7 +137,7 @@ public class MappedBytes extends AbstractBytes<Void> implements Closeable {
         super.writeCheckOffset(offset, adding);
     }
 
-    private void acquireNextByteStore(long offset) {
+    private synchronized void acquireNextByteStore(long offset) {
         BytesStore oldBS = bytesStore;
         try {
             bytesStore = (BytesStore) mappedFile.acquireByteStore(offset);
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/bytes/MappedBytes.java
```diff
@@ -137,7 +137,7 @@ protected void writeCheckOffset(long offset, long adding) throws BufferOverflowE
         super.writeCheckOffset(offset, adding);
     }
 
-    private void acquireNextByteStore(long offset) {
+    private synchronized void acquireNextByteStore(long offset) {
         BytesStore oldBS = bytesStore;
         try {
             bytesStore = (BytesStore) mappedFile.acquireByteStore(offset);
```
scottkidder: Thread A acquires Next
Thread A assigns Next to bytesStore
Thread B assigns bytesStore(Next) to oldBS
Thread A releases oldBS
Thread B releases oldBS(Next)

Next.refCount == 0 and Next is released

leventov:
Maybe this should be synchronized externally to Chronicle Bytes (on usage site)?

RobAustin:
Thanks for your pull request, but please correct me if I am wrong, my assumption is that these changes address and issue that you are seeing around reference counting. 

I'm currently working on a rewrite of the way that we do reference counting, by passing a unique instance, to release/reserve, I will call this instance the ‘owner’ so for example :

- release(owner)
- reserve(owner)

This way we can check that the calls to release() and reserve() are balanced and hopefully fix the reference counting issues.

I feel that by adding synchronised this is a work around, rather than addressing the underlying problem ( but I could be wrong about this so I apologise in advance, if I have got the wrong end of the stick )

PR #26, opened 2016-11-03
More bytes fixes

leventov:
See commit messages for details.


peter-lawrey:
Will retest other modules as well.

Issue #27, opened 2016-11-03
In NativeBytes.resize(), should we silently consume IllegalStateException?


leventov:
Why https://github.com/OpenHFT/Chronicle-Bytes/blob/master/src/main/java/net/openhft/chronicle/bytes/NativeBytes.java#L159-L161?

In case of ISE, it seems the data will be vanished (or corrupted), and we silently continue working. Shouldn't it better just leave ISE for the caller (and then `this.bytesStore` won't be replaced and released)?

peter-lawrey:
At one point I made a few exception checked (see a sub module in core) to
ensure unchecked exceptions were handled correctly. In this case the catch
might have been added incorrectly.  If so, it should be wrapped with
AssertionError if it should never happen.

Need to review it.

hft-team-city:
Released in [Chronicle-Bytes-2.20.101](https://github.com/OpenHFT/Chronicle-Bytes/releases/chronicle-bytes-2.20.101), [BOM-2.20.134](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.20.134)

PR #28, opened 2016-11-04
More bug fixes

leventov:
mostly about incrementing counter by 1 while copying/reading/writing/comparing 8 bytes

Also, what is about #27?

Also, could you please make a Chronicle Core/Bytes release after merging this PR and fixing/reviewing #27?


peter-lawrey:
Thank you for the bug fixes.

What is the change to uncheckedReadUnsignedByte? Or it is about readability/debugging?

leventov:
@peter-lawrey from the surrounding code it seems to expect `BufferOverflowException`, if so increment readPosition only after the read on bytesStore is successful (didn't throw BufferOverflowException)

peter-lawrey:
I though post increment always occurred after the expression was evaluated,
but this is clearer in any case.

PR #42, opened 2017-09-18
Update pom.xml

MartyIX:
https://issues.apache.org/jira/projects/SUREFIRE/versions/12340297 - newer version of Apache Surefire plugin that fixes some more Java 9 issues


epickrram:
I've updated the surefire plugin version in the OpenHFT/java-parent-pom project, which is imported by all the other projects. This helps to keep everything running on the same version for consistency.

PR #54, opened 2018-02-14
assertion in toString

tshymbra:
following line asserts when doing toString on a MappedBytesStore object
assert bytes.start() <= bytes.readPosition();

java.lang.AssertionError
	at net.openhft.chronicle.bytes.BytesInternal.toString(BytesInternal.java:1067)
	at net.openhft.chronicle.bytes.BytesInternal.toString(BytesInternal.java:1021)
	at net.openhft.chronicle.bytes.NativeBytesStore.toString(NativeBytesStore.java:538)

Does MappedBytesStore need readPosition() override that returns its start instead of 0?


dpisklov:
Fixed in 53a9f6e212

PR #95, opened 2019-02-14
Polish: convert to OpenHFT @NotNull and @Nullable

JanStureNielsen:
Based on @peter-lawrey answer to this [SO question](https://stackoverflow.com/questions/54677678/2-x-notnull-too-much), this PR convert the `NotNull` and `Nullable` annotations to the those provided in OpenHFT and removes the IntelliJ annotation dependency.


MartyIX:
@JanStureNielsen I'm just curious: Does IntelliJ IDEA really show correct hints after the PR is applied? (Is nullable variable yellowed out when it is not checked?)

RobAustin:
I thought that the import net.openhft.chronicle.core.annotation.NotNull; can throw a Null Pointer Exception, So I think this change is a change of behaviour. I don't think it is worth making this change to use the chronicle @NotNull exception

MartyIX:
https://github.com/OpenHFT/Chronicle-Core/blob/master/src/main/java/net/openhft/chronicle/core/annotation/NotNull.java is a custom annotation. It cannot throw an exception. 

PS: I have compiled some simple notes on custom annotations when I was researching it:

```
* Java annotations:
    * @Documented - java.lang.annotation.Documented;
        *  *@Documented annotation indicates that whenever the specified annotation is used those elements should be documented using the Javadoc tool. (By default, annotations are not included in Javadoc.)*
    * @Retention - java.lang.annotation.Retention;
        * @Retention annotation specifies how the marked annotation is stored:
            * RetentionPolicy.SOURCE – The marked annotation is retained only in the source level and is ignored by the compiler.
                * *Won't appear in the decompiled class*
            * RetentionPolicy.CLASS – The marked annotation is retained by the compiler at compile time (i.e. it is stored in .class file), but is ignored by the Java Virtual Machine (JVM).
                * *Appear in the decompiled class, but can't be inspected at run-time with reflection with getAnnotations()*
            * RetentionPolicy.RUNTIME – The marked annotation is retained by the JVM so it can be used by the runtime environment.
                * *Appear in the decompiled class, and can be inspected at run-time with reflection with getAnnotations()*
    * @Target - java.lang.annotation.Target;
    * Resources:
        * https://stackoverflow.com/questions/3107970/how-do-different-retention-policies-affect-my-annotations
```

JanStureNielsen:
I don't use IntelliJ, @MartyIX, so I cannot test if that IDE can use non-native `NotNull` and `Nullable` annotations -- searching around it's not very clear. Any IntelliJ IDEA users willing to take a look?

MartyIX:
Let me check

JanStureNielsen:
In the SO issue, I listed the (current) definitions of these annotations, @RobAustin -- the only difference between the two is the default value...so if there is a change in behavior it should be limited to that.

RobAustin:
One reason that this is a bad idea is we often use IntelliJ to generate the @NotNull annotations. and It will end up re-adding them to methods that already have the chronicle @NotNull annotation, for this reason alone I'm going to reject this pull request. Sorry.

MartyIX:
@JanStureNielsen Just a note: There is https://www.jetbrains.com/help/idea/parametersarenonnullbydefault-annotation.html and I found it great when one wants to stick with "parameters are non-null by default" style. I have used this in a project of mine and I was super happy with that. One can come up with a custom annotation that includes non-null default to return values too.

MartyIX:
@RobAustin The `@null`/`@nullable` annotations are messy in Chronicle repositories. It would be great to unify it somehow. General trend (seems to be) is to move to "NonNull by default". For example, Spring Boot adopted that a few months ago AFAIK and their experience was that it allowed them to remove a lot of null-check IF statements. It really makes life easier - i.e. it's cheaper.

What do you mean by "IntelliJ generates @NotNull annotations"? It doesn't do that for me or it is turned off in my Idea.

JanStureNielsen:
I see the convenience of `ParametersAreNonnullByDefault`, @MartyIX , though I prefer specifying this explicitly.

I don't like folding editor or IDE-specific elements into my own projects - it a slippery slope. While some IDEs provide helpful features, like you pointed out @RobAustin, in my view this is not a robust reason...of course, I'm likely biased as an Emacs user! ;-D

In my view, a useful library is generally well-worth the additional dependency. In this particular case, however, the utility of this annotation library is very limited existing project annotations serve the same purpose; adding a dependency seems inappropriate to me. Alternatively, if an external library is preferred, then the project should not provide similar annotations.

So, should the Chronicle-Core `NotNull` and `Nullable` annotations be removed?

MartyIX:
@JanStureNielsen `ParametersAreNonnullByDefault` is not an IDE-specific, it should work in Emacs too. And as such it should be a robust solution.

JanStureNielsen:
Yes; I'm referring to IDE features like flagging and code-modifications. I see the convenience of `@ParametersAreNonnullByDefaul`; I just prefer being explicit.

JanStureNielsen:
As an example of the "messiness" of having duplicate annotations in play, [this method's return is both `@Nullable` and `@NotNull`](https://github.com/OpenHFT/Chronicle-Bytes/blob/master/src/main/java/net/openhft/chronicle/bytes/BytesInternal.java#L1721). :frowning:

JerryShea:
That's not good. I've fixed that specific example.

JanStureNielsen:
Here's another example of (I suspect) [tooling creating a mess](https://github.com/OpenHFT/Chronicle-Bytes/blob/master/src/main/java/net/openhft/chronicle/bytes/BytesStore.java#L629).

RobAustin:
Idea can generate the annotations via Analyze | Infer Nullity. See. https://blog.jetbrains.com/idea/2010/12/auto-infer-nullablenotnull-annotations/

JanStureNielsen:
Hi @RobAustin -- according to [IDEA-64677](https://youtrack.jetbrains.net/issue/IDEA-64677) IntelliJ can be configured to use Chronicle-Core's `@NotNull` and `@Nullable` annotations. If this can be confirmed, would you reconsider this PR?

dpisklov:
@JanStureNielsen In internal discussion with @peter-lawrey and @RobAustin it was agreed that we will delete OpenHFT @Nullable/@NotNull annotations and will only use intellij's ones.

JanStureNielsen:
Sounds good, @dpisklov -- want a PR for that?

dpisklov:
@JanStureNielsen sure thing

RobAustin:
Thanks, and if you are willing to do it, In addition, peter has requested that we update every project to include the IntelliJ NotNull plugin ( I think it is this, but I will have to check ) https://github.com/VladRassokhin/intellij-annotations-instrumenter-maven-plugin <https://github.com/VladRassokhin/intellij-annotations-instrumenter-maven-plugin>, It makes sense if this was added to the <artifactId>java-parent-pom</artifactId>

Then all the projects will have to be updated to use the new java parent pom

RobAustin:
Once the Java Parent POM is updated ( via pull request ) I can release the new parent pom

JanStureNielsen:
@RobAustin -- Vlad's version looks discontinued; [this successor instead](https://github.com/osundblad/intellij-annotations-instrumenter-maven-plugin), perhaps?

RobAustin:
Yep - I was looking through it as your comment came in, and agree that using this makes more sense - https://github.com/osundblad/intellij-annotations-instrumenter-maven-plugin

PR #155, opened 2020-12-06
Feature/mapped file multi thread resize

JerryShea:
Client reported issue https://github.com/ChronicleEnterprise/Chronicle-Queue-Enterprise/issues/156

I don't think we need the complexity of WeakReferences when dealing with synchronizing over FileLocks, as we can use the interned canonical path instead, and this makes the code easier to reason about.

And makes it classloader-safe - if MappedFile were loaded in >1 classloaders then the static FILE_LOCKS can not be relied upon. String#intern is JVM-wide.


JerryShea:
See also https://github.com/OpenHFT/Chronicle-Queue/issues/772

glukos (approved):

PR #154, opened 2020-12-05
Initialize NULL collection fields when deserializing

dkhamitov:
It seems natural to have `null` collection fields initialized when deserialization takes place. E.g. this is already done in other `...FieldAccess` implementations. `BytesFieldAccess`, `MapFieldAccess`, to name a few. Of course, in some scenarios, users can still benefit from defaulting these fields to empty collections.

PR #160, opened 2021-01-19
Initialize NULL collection fields when deserializing

dkhamitov:
See #154 which was closed after migration to `develop`/`ea` branching model. As discussed with @RobAustin, I'm reopening my PR to the `develop` branch with my comment copy-pasted from #154.

It seems natural to have `null` collection fields initialized when deserialization takes place. E.g. this is already done in other `...FieldAccess` implementations. `BytesFieldAccess`, `MapFieldAccess`, to name a few. Of course, in some scenarios, users can still benefit from defaulting these fields to empty collections.

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/BytesMarshaller.java
+++ b/src/main/java/net/openhft/chronicle/bytes/BytesMarshaller.java
@@ -439,7 +439,11 @@ throws IllegalStateException {
                         field.set(o, null);
                     return;
                 }
-                c.clear();
+                if (c == null) {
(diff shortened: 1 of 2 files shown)


RobAustin (approved):

RobAustin:
I’m going to hold off merging this for now as we are in a code freeze period.

peter-lawrey:
We need to check that when a collection is `null` when serialized it should be `null` when deserialized.

dkhamitov:
I believe this PR can be closed as the capability in question has been introduced in https://github.com/OpenHFT/Chronicle-Bytes/issues/196

Issue #192, opened 2021-04-23
ARMMemory initialization might cause deadlock


minborg:
Referencing the subclass `ARMMemory` from its superclass `UnsafeMemory` initializer might lead to class loading deadlock. 

I suggest moving `ARMMemory` to a separate (internal) class.

PR #194, opened 2021-05-08
#192 Ensure VanillaBytes don't require 3x capacity. 

### LINKED ISSUE #193: Add a test
Add a test to Chronicle-Bytes for this use case e.g.

        final String SYMBOL_STR = "LCOM1";
        final Bytes SYMBOL = Bytes.allocateDirect(SYMBOL_STR.length());
        SYMBOL.clear();
        SYMBOL.append(SYMBOL_STR);

jkolobok:


src/main/java/net/openhft/chronicle/bytes/VanillaBytes.java
```diff
@@ -600,7 +600,7 @@ public int byteCheckSum(int start, int end)
         @Override
         public Bytes<Underlying> appendUtf8(char[] chars, int offset, int length)
                 throws BufferOverflowException, IllegalStateException, BufferUnderflowException, IllegalArgumentException {
-            ensureCapacity(writePosition() + length * 3L);
```
JerryShea, line 603: Nice fix. `length` is the length in chars not bytes, although `offset` is in bytes I believe. Maybe we need to specify this in javadoc?

JerryShea (approved):

src/main/java/net/openhft/chronicle/bytes/VanillaBytes.java
jkolobok: yeah. you are right. it doesn't look clear... Thinking...

peter-lawrey:
If there isn't plenty of space i.e. 3 * length, we can find the length it will encode it as with `AppendableUtil.findUtf8Length(chars)` and this can be used to ensureCapacity instead.

src/main/java/net/openhft/chronicle/bytes/VanillaBytes.java
JerryShea: May need to make the same fix to net.openhft.chronicle.bytes.UncheckedNativeBytes#appendUtf8

JerryShea (approved):

hft-team-city:
Released in [Chronicle-Bytes-2.21ea48](https://github.com/OpenHFT/Chronicle-Bytes/releases/chronicle-bytes-2.21ea48), [BOM-2.21ea131](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.21ea131)

PR #227, opened 2021-10-18
Override hashCode to avoid reading data

alamar:
fixes ChronicleEnterprise/Chronicle-FIX#838

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/internal/CommonMappedBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/internal/CommonMappedBytes.java
@@ -592,6 +592,21 @@ public abstract class CommonMappedBytes extends MappedBytes {
     }
 
     @Override
+    // Avoid reading contents as it may be empty or point to unmapped region.
+    public int hashCode() {
+        return mappedFile.hashCode();
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        CommonMappedBytes that = (CommonMappedBytes) o;
+        return mappedFile.equals(that.mappedFile);
+    }
+
+    @Override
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/bytes/internal/CommonMappedBytes.java
```diff
@@ -591,6 +591,12 @@ public String toString() {
                 '}';
     }
 
+    @Override
+    // Avoid reading contents as it may be empty or point to unmapped region.
+    public int hashCode() {
+        return mappedFile.hashCode();
```
JerryShea, line 597: I had not realised that the reference accounting called hashCode(). Why was it crashing before? Were we not pointing to valid memory?

JerryShea (approved):

src/main/java/net/openhft/chronicle/bytes/internal/CommonMappedBytes.java
minborg: This is the same problem as for ByteBuffer. It is questionable to have hashCode() and equals() implementations like these for memory regions. Make sure equals() is overridden in the same way as hashCode() !

alamar: I have added equals() @minborg

minborg:
This is a behavioral change. What is the general principle when it comes to equal(), hashcode() for these memory segments? We should apply the same principle for all memory segment types. Perhaps we should have a short meeting about this. 

The current solution is fundamentally flawed and buggy. For example, even though we can use a `long` for addressing, hashing is limited to 2^31 bytes. It is likely common practice to use Bytes as key in a Map<CharSequence, ?>. Calling hashCode on large memory segments is very slow and would have impacts on caching etc. So, many conflicting requirements here.

alamar:
It would indeed be nice to have a short meeting.

Issue #228, opened 2021-10-21
Improve Javadoc for Bytes.write methods


JerryShea:
Following a customer query, we should improve documentation for these methods - in relation to how readPosition() and writePosition() are treated. When we specify the writeOffset, the writePosition() will not be changed, and likewise for readOffset.

i.e. in RandomDataOuput:
```
R write(long writeOffset, RandomDataInput bytes, long readOffset, long length) throws
R write(long offsetInRDO, @NotNull byte[] bytes) throws ...
```

Related to #140

hft-team-city:
Released in [Chronicle-Bytes-2.22ea9](https://github.com/OpenHFT/Chronicle-Bytes/releases/chronicle-bytes-2.22ea9), [BOM-2.22ea48](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.22ea48)

PR #229, opened 2021-10-21
add javadoc #228

JerryShea:


nicktindall (approved):

minborg:
Generally, we should make sure JavaDocs meets certain criteria and adheres to certain styles. At the very least, it should pass Java 11/17 which is more formal (e.g. detecting missing tags). I will write a short guide.

PR #232, opened 2021-10-21
Removed SonarQube warnings

RobAustin:
Add missing null check

Check for null

Remove redundant && operations

Add assertion for null

Use explicit long in calculation

Remove unused private method

Remove another unused private method

Remove unused private method and rename other method

Fix tests that were incorrect

Remove redundant override

Consolidate to single return

Fix problems in AbstractInterner

Add back override that changed the type

Add assertions to tests

Revert renamed field

Add assertions to tests

Reduce warnings

Deduplicate strings

Fix more warnings

Fix more warnings

Fix warnings

Fix even more warnings

Fix last batch of major issues

Fix remaining issues easily fixable

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -134,7 +134,8 @@
                         <configuration>
                             <referenceVersion>2.22ea3</referenceVersion>
                             <artifactsURI>https://teamcity.chronicle.software/repository/download</artifactsURI>
-                            <binaryCompatibilityPercentageRequired>83.6</binaryCompatibilityPercentageRequired>
+                            <!--Abstract classes got protected constructors-->
+                            <binaryCompatibilityPercentageRequired>99.9</binaryCompatibilityPercentageRequired>
                         </configuration>
                     </execution>
                 </executions>
--- a/pom.xml
+++ b/pom.xml
@@ -134,7 +134,7 @@
                         <configuration>
                             <referenceVersion>2.22ea3</referenceVersion>
                             <artifactsURI>https://teamcity.chronicle.software/repository/download</artifactsURI>
-                            <!--                            <binaryCompatibilityPercentageRequired>92</binaryCompatibilityPercentageRequired>-->
+                            <binaryCompatibilityPercentageRequired>83.6</binaryCompatibilityPercentageRequired>
                         </configuration>
                     </execution>
                 </executions>
--- a/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
@@ -38,18 +38,18 @@ import java.nio.ByteBuffer;
 /**
  * Abstract representation of Bytes.
  *
- * @param <U> Underlying type
+ * @param <Underlying> Underlying type
  */
 @SuppressWarnings("rawtypes")
-public abstract class AbstractBytes<U>
+public abstract class AbstractBytes<Underlying>
         extends AbstractReferenceCounted
-        implements Bytes<U> {
+        implements Bytes<Underlying> {
     private static final boolean BYTES_BOUNDS_UNCHECKED = Jvm.getBoolean("bytes.bounds.unchecked", false);
     // used for debugging
     @UsedViaReflection
     private final String name;
     @NotNull
-    protected BytesStore<Bytes<U>, U> bytesStore;
+    protected BytesStore<Bytes<Underlying>, Underlying> bytesStore;
     protected long readPosition;
     protected long writePosition;
     protected long writeLimit;
@@ -58,12 +58,12 @@ public abstract class AbstractBytes<U>
     private boolean lenient = false;
     private boolean lastNumberHadDigits = false;
 
-    AbstractBytes(@NotNull BytesStore<Bytes<U>, U> bytesStore, long writePosition, long writeLimit)
+    AbstractBytes(@NotNull BytesStore<Bytes<Underlying>, Underlying> bytesStore, long writePosition, long writeLimit)
             throws IllegalStateException {
         this(bytesStore, writePosition, writeLimit, "");
     }
 
-    AbstractBytes(@NotNull BytesStore<Bytes<U>, U> bytesStore, long writePosition, long writeLimit, String name)
+    AbstractBytes(@NotNull BytesStore<Bytes<Underlying>, Underlying> bytesStore, long writePosition, long writeLimit, String name)
             throws IllegalStateException {
         super(bytesStore.isDirectMemory());
         this.bytesStore(bytesStore);
@@ -95,7 +95,7 @@ public abstract class AbstractBytes<U>
 
     @NotNull
     @Override
-    public Bytes<U> compact()
+    public Bytes<Underlying> compact()
             throws IllegalStateException {
         long start = start();
         long readRemaining = readRemaining();
@@ -113,7 +113,7 @@ public abstract class AbstractBytes<U>
 
     @Override
     @NotNull
-    public Bytes<U> clear()
+    public Bytes<Underlying> clear()
             throws IllegalStateException {
         long start = start();
         readPosition = start;
@@ -124,7 +124,7 @@ public abstract class AbstractBytes<U>
 
     @NotNull
     @Override
-    public Bytes<U> clearAndPad(long length)
+    public Bytes<Underlying> clearAndPad(long length)
             throws BufferOverflowException {
         if ((start() + length) > capacity()) {
             throw newBOERange(start(), length, "clearAndPad failed. Start: %d + length: %d > capacity: %d", capacity());
@@ -165,7 +165,7 @@ public abstract class AbstractBytes<U>
 
     @Nullable
     @Override
-    public U underlyingObject() {
+    public Underlying underlyingObject() {
         return bytesStore.underlyingObject();
     }
 
@@ -229,7 +229,7 @@ public abstract class AbstractBytes<U>
 
     @NotNull
     @Override
-    public Bytes<U> readPosition(long position)
+    public Bytes<Underlying> readPosition(long position)
             throws BufferUnderflowException, IllegalStateException {
         if (position < start()) {
             throw new DecoratedBufferUnderflowException(String.format("readPosition failed. Position: %d < start: %d", position, start()));
@@ -244,7 +244,7 @@ public abstract class AbstractBytes<U>
 
     @NotNull
     @Override
-    public Bytes<U> readLimit(long limit)
+    public Bytes<Underlying> readLimit(long limit)
             throws BufferUnderflowException {
         if (limit < start())
             throw limitLessThanStart(limit);
@@ -266,7 +266,7 @@ public abstract class AbstractBytes<U>
 
     @NotNull
     @Override
-    public Bytes<U> writePosition(long position)
+    public Bytes<Underlying> writePosition(long position)
             throws BufferOverflowException {
         if (position > writeLimit())
             throw writePositionTooLarge(position);
@@ -293,7 +293,7 @@ public abstract class AbstractBytes<U>
 
     @NotNull
     @Override
-    public Bytes<U> readSkip(long bytesToSkip)
+    public Bytes<Underlying> readSkip(long bytesToSkip)
             throws BufferUnderflowException, IllegalStateException {
         if (lenient) {
             bytesToSkip = Math.min(bytesToSkip, readRemaining());
@@ -314,7 +314,7 @@ public abstract class AbstractBytes<U>
 
     @NotNull
     @Override
-    public Bytes<U> writeSkip(long bytesToSkip)
+    public Bytes<Underlying> writeSkip(long bytesToSkip)
             throws BufferOverflowException, IllegalStateException {
         final long writePos = writePosition();
         writeCheckOffset(writePos, bytesToSkip);
@@ -324,7 +324,7 @@ public abstract class AbstractBytes<U>
 
     @NotNull
     @Override
-    public Bytes<U> writeLimit(long limit)
+    public Bytes<Underlying> writeLimit(long limit)
             throws BufferOverflowException {
         if (limit < start()) {
             throw writeLimitTooSmall(limit);
@@ -544,7 +544,7 @@ public abstract class AbstractBytes<U>
 
     @NotNull
     @Override
-    public Bytes<U> writeByte(long offset, byte i)
+    public Bytes<Underlying> writeByte(long offset, byte i)
             throws BufferOverflowException, IllegalStateException {
         writeCheckOffset(offset, 1);
         bytesStore.writeByte(offset, i);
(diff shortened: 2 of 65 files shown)


pom.xml
```diff
@@ -134,7 +134,7 @@
                         <configuration>
                             <referenceVersion>2.22ea3</referenceVersion>
                             <artifactsURI>https://teamcity.chronicle.software/repository/download</artifactsURI>
-                            <!--                            <binaryCompatibilityPercentageRequired>92</binaryCompatibilityPercentageRequired>-->
+                            <binaryCompatibilityPercentageRequired>83.6</binaryCompatibilityPercentageRequired>
```
RobAustin: why ?

src/main/java/net/openhft/chronicle/bytes/StreamingDataInput.java
```diff
@@ -482,7 +482,7 @@ default S unsafeRead(long address, int length) {
         if (isDirectMemory()) {
             long src = addressForRead(readPosition());
             readSkip(length);
-            MEMORY.copyMemory(src, address, length);
+            UnsafeMemory.copyMemory(src, address, length);
```
RobAustin, line 485: should we not use MEMORY rather than UnsafeMemory

src/main/java/net/openhft/chronicle/bytes/ref/BinaryIntArrayReference.java
```diff
@@ -311,11 +314,11 @@ public ByteableIntArrayValues capacity(long arrayLength)
         throwExceptionIfClosedInSetter();
 
         BytesStore bytesStore = bytesStore();
-        long length = sizeInBytes(arrayLength);
+        long len = sizeInBytes(arrayLength);
```
RobAustin, line 317: why the name change ?

RobAustin:
@minborg  - per wrote, this I reviewed it, it looks good to me.

minborg:
So, there are some false positives in the binary compatibility checker. Hence, I have updated the percentage accordingly. 

Unfortunately, the tool did not detect that some generics types got new names (due to regex rules triggering), and so these were also flagged falsely. 

I also changed the visibility of abstract constructors to protected. The tool picked up that too but in reality, this would not have an impact. Here are the details:

image

src/main/java/net/openhft/chronicle/bytes/ref/BinaryIntArrayReference.java
minborg: So, oftentimes there are properties `length`, `memory` and `address` and I have opted to rename these to avoid shadowing which triggers issues in SQ.

src/main/java/net/openhft/chronicle/bytes/StreamingDataInput.java
minborg: While it is possible to access static members from a class instance, it’s considered by most to be misleading because it implies to the readers of your code that there’s an instance of the member per class instance.

pom.xml
minborg: See comments below. The checker picks up some changes which it shouldn't.

src/main/java/net/openhft/chronicle/bytes/BytesStore.java
```diff
@@ -517,7 +519,7 @@ default boolean startsWith(char c)
      */
     default boolean contentEquals(@Nullable BytesStore bytesStore)
             throws IllegalStateException {
-        return BytesInternal.contentEqual(this, bytesStore);
+        return bytesStore != null && BytesInternal.contentEqual(this, bytesStore);
```
JerryShea, line 522: nice

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
```diff
@@ -760,19 +760,19 @@ public String read8bit()
...
     @Override
-    public <ACS extends Appendable & CharSequence> long readUtf8(long offset, @NotNull ACS sb)
+    public <C extends Appendable & CharSequence> long readUtf8(long offset, @NotNull C sb)
             throws IORuntimeException, IllegalArgumentException, BufferUnderflowException, ArithmeticException, IllegalStateException {
         return base.readUtf8(offset, sb);
     }
 
     @Override
-    public <ACS extends Appendable & CharSequence> long readUtf8Limited(long offset, @NotNull ACS sb, int maxUtf8Len)
+    public <C extends Appendable & CharSequence> long readUtf8Limited(long offset, @NotNull C sb, int maxUtf8Len)
```
JerryShea, line 775: why change ACS->C and Underlying->U? They seem to be the biggest diffs

src/main/java/net/openhft/chronicle/bytes/VanillaBytes.java
```diff
@@ -449,8 +451,10 @@ public VanillaBytes appendUtf8(@NotNull CharSequence str)
             final long address = bytesStore.address + bytesStore.translate(offset);
             @Nullable final Memory memory = bytesStore.memory;
 
-            if (memory == null)
+            if (memory == null) {
                 bytesStore.throwExceptionIfReleased();
+                throw new NullPointerException("byteStore.memory is null.");
```
JerryShea, line 456: this is behaviour change?

src/main/java/net/openhft/chronicle/bytes/algo/BytesStoreHash.java
```diff
@@ -30,13 +30,13 @@
 @SuppressWarnings("rawtypes")
 public interface BytesStoreHash<B extends BytesStore> extends ToLongFunction<B> {
     static long hash(@NotNull BytesStore b) {
-        return b.isDirectMemory() && b.isDirectMemory()
```
JerryShea, line 33: excellent

src/test/java/net/openhft/chronicle/bytes/BytesTest.java
```diff
@@ -597,7 +599,9 @@ public void testMove2B() {
         assertEquals("Hlo o World", b.toString());
         postTest(b);
         BackgroundResourceReleaser.releasePendingResources();
-        b.bytesStore().move(3, 5, 3);
+        assertThrows(IllegalStateException.class, () ->
```
JerryShea, line 618: like

src/test/java/net/openhft/chronicle/bytes/MappedBytesTest.java
```diff
@@ -88,6 +91,9 @@ public void testMappedFileSafeLimitTooSmall2()
             for (int i = 0; i < 5; i++) {
                 bytesR.write(data);
             }
+        } catch (Exception e) {
+            e.printStackTrace();
```
JerryShea, line 95: won't Sonar complain about printStackTrace() exposing implementation details or is that only for non-test code?

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
JerryShea: Call me picky, but I think I would prefer to see one PR without these renames then another with just the renames - it feels like we are dealing with 2 cross-cutting concerns. I don't feel that strongly - I'll approve - just trying to reduce the cognitive load on the reviewer

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -602,11 +609,11 @@ static void parseUtf82(@NotNull StreamingDataInput bytes, @NotNull Appendable ap
                     count += utf ? 2 : 1;
                     if (count > length)
                         throw new UTFDataFormatRuntimeException(
-                                "malformed input: partial character at end");
+                                MALFORMED_INPUT_PARTIAL_CHARACTER_AT_END);
                     int char2 = bytes.readUnsignedByte();
                     if ((char2 & 0xC0) != 0x80)
                         throw new UTFDataFormatRuntimeException(
-                                "malformed input around byte " + count + " was " + char2);
+                                MALFORMED_INPUT_AROUND_BYTE + count + WAS + char2);
```
JerryShea, line 616: maybe should printf?

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -1478,7 +1485,7 @@ public static void appendDecimal(@NotNull RandomDataOutput out, long num, long o
             out.writeUnsignedByte(offset++, '.');
             while (decimalPlaces-- > digits)
                 out.writeUnsignedByte(offset++, '0');
-            out.write(offset++, numberBuffer, endIndex, digits);
+            out.write(offset, numberBuffer, endIndex, digits);
```
JerryShea, line 1488: behaviour change?

JerryShea: ah, no its not

JerryShea (approved):
Looks good. A few minor questions/comments

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -308,17 +314,17 @@ public void uncheckedReadSkipBackOne() {
...
     @Override
-    public Bytes<Underlying> writeSkip(long bytesToSkip)
+    public Bytes<U> writeSkip(long bytesToSkip)
             throws BufferOverflowException, IllegalStateException {
-        final long writePosition = writePosition();
-        writeCheckOffset(writePosition, bytesToSkip);
-        uncheckedWritePosition(writePosition + bytesToSkip);
+        final long writePos = writePosition();
+        writeCheckOffset(writePos, bytesToSkip);
+        uncheckedWritePosition(writePos + bytesToSkip);
```
nicktindall, line 321: I like longer variable names, this isn't a good example, but I think abbreviations unnecessarily hurt search-ability

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
nicktindall: I know sonar doesn't like generic params longer than a single char. I'm not sure about that rule, I've found myself just doing it to stop Sonar complaining in the past but arguably it reduces readability?

nicktindall: Probably OK when there's one or two generic params, but beyond that why would you limit yourself to a single char?

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
nicktindall: Or format(..) in this case, the template string would make more sense as a constant than the two separate string

src/main/java/net/openhft/chronicle/bytes/internal/NativeBytesStore.java
```diff
@@ -213,9 +214,9 @@ public void uninit() {
     public void move(long from, long to, long length)
             throws BufferUnderflowException, IllegalStateException {
         if (from < 0 || to < 0) throw new BufferUnderflowException();
-        long address = this.address;
-        if (address == 0) throwException(null);
-        memoryCopyMemory(address + from, address + to, length);
+        long addr = this.address;
+        if (addr == 0) throwException(null);
+        memoryCopyMemory(addr + from, addr + to, length);
```
nicktindall, line 219: Again, I don't think we should use abbreviations

src/main/java/net/openhft/chronicle/bytes/ref/BinaryIntArrayReference.java
nicktindall: Could we use something like `newLength` or `lengthOfThing` or something? I think having `len` and `length` in scope is potentially more hazardous as you could easily use the wrong one. And I don't like abbreviations ;)

nicktindall (approved):

JerryShea (dismissed):
hi @minborg can we hold off merging this until @nicktindall has applied a couple of fixes to Bytes? We are likely to need to backport these fixes to .21

src/main/java/net/openhft/chronicle/bytes/ref/BinaryIntArrayReference.java
minborg: Good point @nicktindall

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
minborg: Point taken @JerryShea. I just ran along and should perhaps divide this work into several PRs.

src/main/java/net/openhft/chronicle/bytes/VanillaBytes.java
minborg: I don't think it is because `memory` is dereferenced in all branches later on and there is no side-effect in between.

src/main/java/net/openhft/chronicle/bytes/algo/BytesStoreHash.java
minborg: initially, I was thinking that this might be some kind of micro-optimization but failed to understand why.

src/test/java/net/openhft/chronicle/bytes/MappedBytesTest.java
minborg: I think the tests are treated less stringent. But let's see if this surface when merged.

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
minborg: I can have a look at that in a separate PR

src/main/java/net/openhft/chronicle/bytes/internal/NativeBytesStore.java
minborg: Let me address this in a separate PR.

RobAustin:
> hi @minborg can we hold off merging this until @nicktindall has applied a couple of fixes to Bytes? We are likely to need to backport these fixes to .21

I don't understand why you can't apply the changes directly to .21  ? The point of having these branches is to let the non-conservative changes go into "ea", while conservative selective changes are only applied to the "21" branch. If there are no technical objections - I would like us to merge this in tomorrow at the latest.

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
JerryShea: Yep @minborg the more I think about this, the more I think we should undo the change to the generic params. TBH it feels like Sonar being unnecessarily opinionated and a rule that we could/should probably turn off. For me the code churn vs benefit does not stack up

pom.xml
minborg: Reverting generic name changes got us back to 99.9%

PR #233, opened 2021-10-21
Enable JU5 while retaining backwards compatibility with JU4

minborg:
This PR introduces the ability to write new tests in either JUnit4 or JUnit5, whichever is deemed most favorable for the issue at hand.

Old JUnit4 test can be migrated to JUnit5 now, later or never.

As @nicktindall indicated, there is a "vintage" dependency that makes this process trivial.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,15 +77,17 @@
         </dependency>
 
         <!-- test dependencies -->
+
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
             <scope>test</scope>
         </dependency>
 
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-simple</artifactId>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <version>5.8.1</version>
             <scope>test</scope>
         </dependency>
 
--- a/src/test/java/net/openhft/chronicle/bytes/BytesTestCommon.java
+++ b/src/test/java/net/openhft/chronicle/bytes/BytesTestCommon.java
@@ -9,6 +9,8 @@ import net.openhft.chronicle.core.threads.ThreadDump;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
 
 import java.util.LinkedHashMap;
 import java.util.Map;
@@ -24,11 +26,13 @@ public class BytesTestCommon {
     protected Map<ExceptionKey, Integer> exceptions;
 
     @Before
+    @BeforeEach
     public void enableReferenceTracing() {
         AbstractReferenceCounted.enableReferenceTracing();
     }
 
     @Before
+    @BeforeEach
     public void threadDump() {
         threadDump = new ThreadDump();
     }
@@ -38,6 +42,7 @@ public class BytesTestCommon {
     }
 
     @Before
+    @BeforeEach
     public void recordExceptions() {
         exceptions = Jvm.recordExceptions();
     }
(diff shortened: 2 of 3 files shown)


src/test/java/net/openhft/chronicle/bytes/BytesTestCommon.java
```diff
@@ -18,16 +20,19 @@
 import static net.openhft.chronicle.core.io.AbstractReferenceCounted.assertReferencesReleased;
 
 public class BytesTestCommon {
+
+    private final Map<Predicate<ExceptionKey>, String> expectedExceptions = new LinkedHashMap<>();
     protected ThreadDump threadDump;
     protected Map<ExceptionKey, Integer> exceptions;
-    private Map<Predicate<ExceptionKey>, String> expectedExceptions = new LinkedHashMap<>();
```
JerryShea, line 23: why change this?

src/test/java/net/openhft/chronicle/bytes/BytesTestCommon.java
```diff
@@ -18,16 +20,19 @@
...
 
     @Before
```
JerryShea, line 28: So the junit-vintage runner (or whatever it is called) will respect `@Before` *and* `@BeforeClass`? 

Do we actually need to change this class at all? Is it only the change to the pom.xml that is required?

src/test/java/net/openhft/chronicle/bytes/VanillaBytesTest.java
```diff
@@ -20,20 +20,20 @@
 package net.openhft.chronicle.bytes;
 
 import org.jetbrains.annotations.NotNull;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
```
JerryShea, line 23: these 2 imports are the only actual changes? (other than cleanup)

pom.xml
```diff
@@ -77,15 +77,17 @@
...
-            <artifactId>junit</artifactId>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
             <scope>test</scope>
         </dependency>
 
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-simple</artifactId>
+            <groupId>org.junit.vintage</groupId>
```
JerryShea, line 88: we should add these version numbers to third-party-bom now and release it

minborg: Already added. :-)

src/test/java/net/openhft/chronicle/bytes/VanillaBytesTest.java
minborg: Yes

nicktindall (approved):

src/test/java/net/openhft/chronicle/bytes/BytesTestCommon.java
```diff
@@ -18,16 +20,19 @@
...
+    @BeforeEach
     public void enableReferenceTracing() {
         AbstractReferenceCounted.enableReferenceTracing();
     }
 
     @Before
+    @BeforeEach
```
nicktindall, line 35: Maybe we should just leave one annotation in?

src/test/java/net/openhft/chronicle/bytes/BytesTestCommon.java
minborg: So, I just moved the declaration upwards so `final` fields are declared above mutable fields.

src/test/java/net/openhft/chronicle/bytes/BytesTestCommon.java
minborg: Vintage will pickup `@Before` and the normal JUnit5 will pickup `@BeforeEach`. This means that this base class can be used by ether version.

src/test/java/net/openhft/chronicle/bytes/BytesTestCommon.java
minborg: We would like the base class to function for either type. With this, we can subclass using both JUnit4 and JUnit5.

src/test/java/net/openhft/chronicle/bytes/BytesTestCommon.java
JerryShea: ah of course - cool

PR #234, opened 2021-10-22
Bytes optimizations from client (WIP)

nicktindall:
Just a PR to share what I have so far.

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
@@ -1015,6 +1015,24 @@ public abstract class AbstractBytes<Underlying>
         return this;
     }
 
+    @Override
+    public int read(@NotNull byte[] bytes, int off, int len) throws BufferUnderflowException, IllegalStateException {
+        long remaining = readRemaining();
+        if (remaining <= 0)
+            return -1;
+        final int totalToCopy = (int) Math.min(len, remaining);
+        int remainingToCopy = totalToCopy;
+        int currentOffset = off;
+        while (remainingToCopy > 0) {
+            int currentBatchSize = Math.min(remainingToCopy, safeCopySize());
+            long offsetInRDO = readOffsetPositionMoved(currentBatchSize);
+            bytesStore.read(offsetInRDO, bytes, currentOffset, currentBatchSize);
+            currentOffset += currentBatchSize;
+            remainingToCopy -= currentBatchSize;
+        }
+        return totalToCopy;
+    }
+
     @NotNull
     @Override
     public Bytes<Underlying> write(@NotNull byte[] bytes, int offset, int length)
--- a/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
@@ -1033,6 +1033,11 @@ public abstract class AbstractBytes<Underlying>
         return totalToCopy;
     }
 
+    @Override
+    public long read(long offsetInRDI, byte[] bytes, int offset, int length) throws IllegalStateException {
+        return bytesStore.read(offsetInRDI, bytes, offset, length);
+    }
+
     @NotNull
     @Override
     public Bytes<Underlying> write(@NotNull byte[] bytes, int offset, int length)
--- a/src/main/java/net/openhft/chronicle/bytes/internal/NativeBytesStore.java
+++ b/src/main/java/net/openhft/chronicle/bytes/internal/NativeBytesStore.java
@@ -824,16 +824,7 @@ public class NativeBytesStore<Underlying>
     @Override
     public long read(long offsetInRDI, byte[] bytes, int offset, int length) {
         int len = (int) Math.min(length, readLimit() - offsetInRDI);
-        int i;
-        final long address = this.address + translate(offsetInRDI);
-        for (i = 0; i < len - 7; i += 8)
-            UnsafeMemory.unsafePutLong(bytes, i, memory.readLong(address + i));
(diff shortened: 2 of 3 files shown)


src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -1015,6 +1015,24 @@ protected long prewriteOffsetPositionMoved(long subtracting)
...
+        int currentOffset = off;
+        while (remainingToCopy > 0) {
+            int currentBatchSize = Math.min(remainingToCopy, safeCopySize());
+            long offsetInRDO = readOffsetPositionMoved(currentBatchSize);
+            bytesStore.read(offsetInRDO, bytes, currentOffset, currentBatchSize);
+            currentOffset += currentBatchSize;
+            remainingToCopy -= currentBatchSize;
+        }
+        return totalToCopy;
+    }
```
nicktindall, line 1034: So this delegates to the BytesStore to do the copy in `safeCopySize()` chunks (not sure why it needs to be chunked, was copying that from write). I guess it gives the BytesStore the opportunity to implement an optimized copy.

src/main/java/net/openhft/chronicle/bytes/internal/NativeBytesStore.java
```diff
@@ -824,16 +825,7 @@ public boolean sharedMemory() {
...
-        final long address = this.address + translate(offsetInRDI);
-        for (i = 0; i < len - 7; i += 8)
-            UnsafeMemory.unsafePutLong(bytes, i, memory.readLong(address + i));
-        if (i < len - 3) {
-            UnsafeMemory.unsafePutInt(bytes, i, memory.readInt(address + i));
-            i += 4;
-        }
-        for (; i < len; i++)
-            UnsafeMemory.unsafePutByte(bytes, i, memory.readByte(address + i));
+        memory.copyMemory(this.address + translate(offsetInRDI), bytes, offset + Unsafe.ARRAY_BYTE_BASE_OFFSET, len);
```
nicktindall: Now this is where it ends up, it had that bug that offset wasn't respected, this change fixes it, but if you look at the `memory.copyMemory` implementation it ends up doing the same thing that the original implementation did

src/main/java/net/openhft/chronicle/bytes/internal/NativeBytesStore.java
```diff
@@ -26,6 +26,7 @@
 import net.openhft.chronicle.core.util.SimpleCleaner;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import sun.misc.Unsafe;
```
JerryShea: shouldn't import this - should use our unsafe UnsafeMemory

src/main/java/net/openhft/chronicle/bytes/internal/NativeBytesStore.java
JerryShea: cool

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
JerryShea: From looking at this some more, I think we may be able to do away with the chunking in this method - but let's review that later today - ah actually, this calls ByteStore.readBytes which does not do chunking. Maybe `net.openhft.chronicle.core.UnsafeMemory#readBytes` should? That is scope creep so let's leave

JerryShea (approved):

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -1015,6 +1015,29 @@ protected long prewriteOffsetPositionMoved(long subtracting)
...
+
+    @Override
+    public long read(long offsetInRDI, @NotNull byte[] bytes, int offset, int length) throws IllegalStateException {
```
JerryShea: needed to add this here. I didn't chunk it

JerryShea (approved):

src/main/java/net/openhft/chronicle/bytes/RandomDataInput.java
```diff
@@ -584,7 +584,7 @@ default boolean compareUtf8(long offset, @Nullable CharSequence other)
         return BytesInternal.toByteArray(this);
     }
 
-    default long read(long offsetInRDI, byte[] bytes, int offset, int length)
+    default long read(long offsetInRDI, @NotNull byte[] bytes, int offset, int length)
```
minborg: `@NotNull` for arrays means that the elements in the array cannot be null. Consequently, this annotation for primitive arrays makes no sense.

src/test/java/net/openhft/chronicle/bytes/BytesTest.java
```diff
@@ -1016,4 +1016,17 @@ public void testAppendReallySmallDouble() {
             assertEquals(d, bytes.parseDouble(), err);
         }
     }
-}
\ No newline at end of file
+
+    @Test
```
minborg, line 1020: +1 for testing

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -1015,6 +1015,29 @@ protected long prewriteOffsetPositionMoved(long subtracting)
         return this;
     }
 
+    @Override
+    public int read(@NotNull byte[] bytes, int off, int len) throws BufferUnderflowException, IllegalStateException {
```
minborg, line 1019: I suggest `byteArray` over `bytes`. After all, we are in the `Bytes` class.

minborg (approved):

PR #237, opened 2021-10-22
Resolve suspicious method UncheckedBytes::write, Fix #230

### LINKED ISSUE #230: Remove unsafe method UncheckedBytes::write
Currently, this method looks like this:

```
    @Override
    public void write(long offsetInRDO, ByteBuffer bytes, int offset, int length)
            throws BufferOverflowException {

    }
```

but it is more likely that it should be like this:

```
    @Override
    public void write(long offsetInRDO, ByteBuffer bytes, int offset, int length)
            throws BufferOverflowException {
        bytesStore.write(offsetInRDO, bytes, offset, length);
    }
```

minborg:
It appears that we missed doing something in this method.

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/UncheckedBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/UncheckedBytes.java
@@ -309,6 +309,7 @@ public class UncheckedBytes<Underlying>
     @Override
     public void write(long offsetInRDO, ByteBuffer bytes, int offset, int length)
             throws BufferOverflowException {
-
+        // Todo: Optimize this
+        super.write(offsetInRDO, bytes, offset, length);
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/bytes/UncheckedBytes.java
```diff
@@ -309,6 +309,7 @@ void append8bit(char[] chars)
     @Override
     public void write(long offsetInRDO, ByteBuffer bytes, int offset, int length)
             throws BufferOverflowException {
-
+        // Todo: Optimize this
```
JerryShea, line 312: great catch. We should have a test? :)

JerryShea (approved):

PR #240, opened 2021-10-25
Make MappedBytes respect the readOnly flag, Fix #239

### LINKED ISSUE #239: MappedBytes does not respect the readOnly flag in its static constructor
As can be seen below, a singleMappedBytes will always be read/write regardless of the provided `readOnly`.

```
    @NotNull
    public static MappedBytes singleMappedBytes(@NotNull final String filename, final long capacity)
            throws FileNotFoundException, IllegalStateException {
        return singleMappedBytes(new File(filename), capacity);
    }

    @NotNull
    public static MappedBytes singleMappedBytes(@NotNull final File file, final long capacity)
            throws FileNotFoundException, IllegalStateException {
        return singleMappedBytes(file, capacity, true);
    }

    @NotNull
    public static MappedBytes singleMappedBytes(@NotNull final File file, final long capacity, boolean readOnly)
            throws FileNotFoundException, IllegalStateException {
        @NotNull final MappedFile rw = MappedFile.ofSingle(file, capacity, false);
        try {
            return new SingleMappedBytes(rw);
        } finally {
            rw.release(INIT);
        }
    }

```

minborg:
This PR fixes `MappedBytes::singleMappedBytes` with the `readOnly` flag while retaining the behaviour for the other overloads.

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/MappedBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/MappedBytes.java
@@ -62,13 +62,13 @@ public abstract class MappedBytes extends AbstractBytes<Void> implements Closeab
     @NotNull
     public static MappedBytes singleMappedBytes(@NotNull final File file, final long capacity)
             throws FileNotFoundException, IllegalStateException {
-        return singleMappedBytes(file, capacity, true);
+        return singleMappedBytes(file, capacity, false);
     }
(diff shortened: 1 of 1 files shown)


minborg:
Please have a look at this so we could merge.

RobAustin (approved):

PR #243, opened 2021-10-26
Throw if resize does not give us enough room

JerryShea:
What the new test tries to do is write a 128K text block to a MappedBytes of 64K chunk size. 
Previously it would crash the VM as the ChunkedMappedBytes would (correctly) detect that the new offset was not inside the BytesStore, acquire a new BytesStore, but still not realise that we were still not inside the new BytesStore.

The only thing we can reasonably do here is throw an exception - we cannot write a message which is larger than the chunk size.

Not ready for merge - for discussion at this stage

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/MappedBytesStore.java
+++ b/src/main/java/net/openhft/chronicle/bytes/MappedBytesStore.java
@@ -104,7 +104,7 @@ public class MappedBytesStore extends NativeBytesStore<Void> {
     @Override
     public boolean inside(long offset, long buffer) {
         // this is correct that it uses the maximumLimit, yes it is different than the method above.
-        return start <= offset && offset + buffer < limit;
+        return start <= offset && offset + buffer <= limit;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/main/java/net/openhft/chronicle/bytes/MappedBytesStore.java
```diff
@@ -104,7 +104,7 @@ public boolean inside(long offset) {
     @Override
     public boolean inside(long offset, long buffer) {
         // this is correct that it uses the maximumLimit, yes it is different than the method above.
-        return start <= offset && offset + buffer < limit;
+        return start <= offset && offset + buffer <= limit;
```
JerryShea, line 107: Otherwise `net.openhft.chronicle.bytes.CopyBytesTest#testCanCopyBytesFromMappedBytes2` tried to get a new chunk when it does not need it

nicktindall (approved):
Certainly better than what was there, but I'd be in favour of making it work.

I don't fully understand the contract though, is the  consumer supposed to know they're using a chunked bytes implementation and know what the size limit is so they can at least handle it gracefully?

I imagine sometimes they're just writing to a BytesStore and aren't aware of implementation specific details? or do they all have queryable write limits?

JerryShea:
We can't make it work - you can't write something longer than the chunk. The consumer has to know. The chunk size is normally pretty big, and we have documented in CQ doco that there are message size limits.

minborg:
I do not know all the details here but I think that the limit should be queryable and/or documented for all libraries including Bytes itself.

peter-lawrey:
I haver taken the view that there is no sane situation where you would want to write a message anywhere close to the overlap size, however I get that it worries people so perhaps we should just fix it even though it has no practical application afaics.

peter-lawrey:
I have added Issue #247 however the advice is unchanged that your overlapping size should be much larger than any message you write, however there shouldn't be critical problem if you do.

PR #246, opened 2021-10-29
JMH benchmarks

JerryShea:


nicktindall (approved):

microbenchmarks/pom.xml
```diff
@@ -36,7 +36,7 @@
                 <groupId>net.openhft</groupId>
                 <artifactId>third-party-bom</artifactId>
                 <type>pom</type>
-                <version>3.19.4</version>
+                <version>3.19.10</version>
```
minborg, line 39: I think there is a 22 available these days

PR #248, opened 2021-10-29
Feature/javadoc

minborg:
This is the first drop of adding JavaDocs and asserting them using JUnit tests. 

It should be noted that this is a work in progress and that, by no means, the process is complete. Instead, I would like to get the reviewers' feedback early before continuing in the same direction with other classes and tests. Hence, the relatively large amount of reviewers being invited.

The area of asserting method parameter invariants for performance applications is grey. We need to strike a balance between correctness and performance. However, in my opinion, there should exist no combination of operations that can make the JVM crash or that can place an object in an illegal state. 

The PR fixes countless of bugs including JVM crashes, unintended side effects for illegal parameters and exceptions not thrown when they should. I have not logged them all.

In total, this PR introduces about 800 new JUnit tests and thousands of new assertions.

This PR will introduce some behavioral changes including (but potentially not limited to):

A) The JVM will not crash when providing null arguments for some methods
B) Mutation of internal state will no longer be made when providing illegal arguments for some methods
C) The order of checking invariants/state will change
D) Some methods will throw a documented exception rather than an accident exception when provided with illegal arguments
E) Numerous undocumented cases are now documented potentially changing (unspecified) behaviour.

The following Bytes implementations are tested:

```
net.openhft.chronicle.bytes.NativeBytes
net.openhft.chronicle.bytes.OnHeapBytes
net.openhft.chronicle.bytes.NativeBytes
net.openhft.chronicle.bytes.VanillaBytes
net.openhft.chronicle.bytes.NativeBytes
net.openhft.chronicle.bytes.internal.ChunkedMappedBytes
net.openhft.chronicle.bytes.internal.SingleMappedBytes
net.openhft.chronicle.bytes.internal.EnbeddedBytes
net.openhft.chronicle.bytes.UncheckedBytes
net.openhft.chronicle.bytes.HexDumpBytes
net.openhft.chronicle.bytes.UncheckedNativeBytes
net.openhft.chronicle.bytes.GuardedNativeBytes
```

As of now, a small number (40) of operations are tested. I expect the number of operations tested to eventually be about 10 times larger only for method parameter invariant checking.

It is expected that there will be a slight performance reduction due to parameter checking. We might have to re-balance correctness vs. performance in a second step following performance testing.

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
@@ -55,12 +55,12 @@ public abstract class AbstractBytes<Underlying>
     private boolean lenient = false;
     private boolean lastNumberHadDigits = false;
 
-    AbstractBytes(@NotNull(exception = NullPointerException.class) BytesStore<Bytes<Underlying>, Underlying> bytesStore, long writePosition, long writeLimit)
+    AbstractBytes(@NotNull BytesStore<Bytes<Underlying>, Underlying> bytesStore, long writePosition, long writeLimit)
             throws IllegalStateException {
         this(bytesStore, writePosition, writeLimit, "");
     }
 
-    AbstractBytes(@NotNull(exception = NullPointerException.class) BytesStore<Bytes<Underlying>, Underlying> bytesStore, long writePosition, long writeLimit, String name)
+    AbstractBytes(@NotNull BytesStore<Bytes<Underlying>, Underlying> bytesStore, long writePosition, long writeLimit, String name)
             throws IllegalStateException {
         super(bytesStore.isDirectMemory());
         this.bytesStore(bytesStore);
@@ -647,7 +647,7 @@ public abstract class AbstractBytes<Underlying>
 
     @Override
     @NotNull
-    public Bytes<Underlying> write(@NotNull(exception = NullPointerException.class) final RandomDataInput bytes)
+    public Bytes<Underlying> write(@NotNull final RandomDataInput bytes)
             throws IllegalStateException, BufferOverflowException {
         assert bytes != this : "you should not write to yourself !";
 
@@ -660,7 +660,7 @@ public abstract class AbstractBytes<Underlying>
 
     @Override
     @NotNull
-    public Bytes<Underlying> write(long offsetInRDO, @NotNull(exception = NullPointerException.class) byte[] bytes, int offset, int length)
+    public Bytes<Underlying> write(long offsetInRDO, @NotNull byte[] bytes, int offset, int length)
             throws BufferOverflowException, IllegalStateException {
         requireNonNull(bytes);
         long remaining = length;
@@ -676,7 +676,7 @@ public abstract class AbstractBytes<Underlying>
     }
 
     @Override
-    public void write(long offsetInRDO, @NotNull(exception = NullPointerException.class) ByteBuffer bytes, int offset, int length)
+    public void write(long offsetInRDO, @NotNull ByteBuffer bytes, int offset, int length)
             throws BufferOverflowException, IllegalStateException {
         requireNonNull(bytes);
         writeCheckOffset(offsetInRDO, length);
@@ -685,7 +685,7 @@ public abstract class AbstractBytes<Underlying>
 
     @Override
     @NotNull
-    public Bytes<Underlying> write(long writeOffset, @NotNull(exception = NullPointerException.class) RandomDataInput bytes, long readOffset, long length)
+    public Bytes<Underlying> write(long writeOffset, @NotNull RandomDataInput bytes, long readOffset, long length)
             throws BufferOverflowException, BufferUnderflowException, IllegalStateException {
 
         requireNonNull(bytes);
@@ -702,7 +702,7 @@ public abstract class AbstractBytes<Underlying>
     }
 
     @Override
-    public @NotNull Bytes<Underlying> write8bit(@NotNull(exception = NullPointerException.class) String text, int start, int length) throws BufferOverflowException, IndexOutOfBoundsException, ArithmeticException, IllegalStateException, BufferUnderflowException {
+    public @NotNull Bytes<Underlying> write8bit(@NotNull String text, int start, int length) throws BufferOverflowException, IndexOutOfBoundsException, ArithmeticException, IllegalStateException, BufferUnderflowException {
         requireNonNull(text); // This needs to be checked or else the JVM might crash
         final long toWriteLength = UnsafeMemory.INSTANCE.stopBitLength(length) + (long) length;
         final long position = writeOffsetPositionMoved(toWriteLength, 0);
@@ -725,12 +725,12 @@ public abstract class AbstractBytes<Underlying>
     }
 
     @Override
-    public long write8bit(long position, @NotNull(exception = NullPointerException.class) BytesStore bs) {
+    public long write8bit(long position, @NotNull BytesStore bs) {
         return bytesStore.write8bit(position, bs);
     }
 
     @Override
-    public long write8bit(long position, @NotNull(exception = NullPointerException.class) String s, int start, int length) {
+    public long write8bit(long position, @NotNull String s, int start, int length) {
         return bytesStore.write8bit(position, s, start, length);
     }
 
@@ -873,7 +873,7 @@ public abstract class AbstractBytes<Underlying>
 
     @NotNull
     @Override
-    public Bytes<Underlying> prewrite(@NotNull(exception = NullPointerException.class) byte[] bytes)
+    public Bytes<Underlying> prewrite(@NotNull byte[] bytes)
             throws BufferOverflowException, IllegalStateException {
         long offset = prewriteOffsetPositionMoved(bytes.length);
         bytesStore.write(offset, bytes);
@@ -882,7 +882,7 @@ public abstract class AbstractBytes<Underlying>
 
     @NotNull
     @Override
-    public Bytes<Underlying> prewrite(@NotNull(exception = NullPointerException.class) BytesStore bytes)
+    public Bytes<Underlying> prewrite(@NotNull BytesStore bytes)
             throws BufferOverflowException, IllegalStateException {
         long offset = prewriteOffsetPositionMoved(bytes.readRemaining());
         bytesStore.write(offset, bytes);
@@ -1023,7 +1023,7 @@ public abstract class AbstractBytes<Underlying>
 
     @NotNull
     @Override
-    public Bytes<Underlying> write(@NotNull(exception = NullPointerException.class) byte[] bytes, int offset, int length)
+    public Bytes<Underlying> write(@NotNull byte[] bytes, int offset, int length)
             throws BufferOverflowException, IllegalStateException, IllegalArgumentException {
 
         if ((length + offset) > bytes.length) {
@@ -1051,7 +1051,7 @@ public abstract class AbstractBytes<Underlying>
 
     @NotNull
     @Override
-    public Bytes<Underlying> writeSome(@NotNull(exception = NullPointerException.class) ByteBuffer buffer)
+    public Bytes<Underlying> writeSome(@NotNull ByteBuffer buffer)
             throws BufferOverflowException, IllegalStateException, BufferUnderflowException {
         int length = (int) Math.min(buffer.remaining(), writeRemaining());
         try {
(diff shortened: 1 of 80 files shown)


minborg:
The reason for adding `@NotNull(exception = NullPointerException.class)` is that it shall be possible to run the tests under IDEA. IDEA instruments classes and throws IEA unless otherwise specified.

RobAustin (approved):

minborg:
Given the number of changes, I am eager to merge this PR ASAP. Although eligible for merge already now, I will hold off until noon UK time on the 1st of November unless any objects before that time.

src/main/java/net/openhft/chronicle/bytes/Bytes.java
```diff
@@ -144,8 +178,30 @@
...
+     * Here is an example of field groups:
+     * <pre>{@code
+     *     static class Padding extends Parent {
+     *         @FieldGroup("p")
+     *         // 128 bytes
+     *         transient long p0, p1, p2, p3, p4, p5, p6, p7, p8, p9, p10, p11, p12, p13, p14, p15;
+     *     }
+     * }</pre>
+     *
+     * @param object    who's fields are to be reflected
```
nicktindall: *whose

src/main/java/net/openhft/chronicle/bytes/Bytes.java
```diff
@@ -207,10 +266,13 @@
      * }</pre>
      *
      * @param byteBuffer to wrap
-     * @return a Bytes which wraps the provided ByteBuffer and is ready for reading.
+     * @return a new Bytes wrapping the provided {@code byteBuffer}
+     * @throws NullPointerException    if the provided {@code byteBuffer} is {@code null}
+     * @throws ReadOnlyBufferException if the provided {@code byteBuffer} is read-only
```
nicktindall, line 266: Should we be throwing a `ReadOnlyBufferException` when getting a read-only buffer? we're wrapping for read aren't we? it seems counter-intuitive that we require a writable buffer for this method

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -53,12 +55,12 @@
     private boolean lenient = false;
     private boolean lastNumberHadDigits = false;
 
-    AbstractBytes(@NotNull BytesStore<Bytes<Underlying>, Underlying> bytesStore, long writePosition, long writeLimit)
+    AbstractBytes(@NotNull(exception = NullPointerException.class) BytesStore<Bytes<Underlying>, Underlying> bytesStore, long writePosition, long writeLimit)
```
JerryShea: I understand why all of these were done, but it does not feel ideal - it adds a lot of extra code, and reduces the signal/noise ration. 

It just feels strange that both Intellij and the [maven plugin we use](https://github.com/osundblad/intellij-annotations-instrumenter-maven-plugin/) throw `IllegalArgumentException` instead of NPE. As you have pointed out [in this issue](https://github.com/osundblad/intellij-annotations-instrumenter-maven-plugin/issues/53) it would seem more useful that this instrumentation should mimic `Objects::requireNonNull` (or our `ObjectUtils::requireNotNull`) and throw a NPE. 

Another approach we could follow is:
* implement [issue 53](https://github.com/osundblad/intellij-annotations-instrumenter-maven-plugin/issues/53) ourselves so we can configure the plugin to throw NPE (in one place - our root-parent-pom)
* (as it does not appear to be possible to configure Intellij's non-null exception globally) so we could turn off the IDE's instrumentation and commit that config to git (for all projects)

src/main/java/net/openhft/chronicle/bytes/Bytes.java
```diff
@@ -667,42 +788,51 @@ static VanillaBytes allocateDirect(@NotNull byte[] bytes)
...
-    static Bytes fromHexString(@NotNull String s) {
+    static Bytes fromHexString(@NotNull(exception = NullPointerException.class) String s) {
         return BytesInternal.fromHexString(s);
     }
 
     /**
-     * Code shared by String and StringBuffer to do searches. The
-     * source is the character array being searched, and the target
+     * Returns the lowest value such that the contents of the provided {@code source } equals the contents of the
+     * provided {@code other } starting at the provided {@code fromSourceOffset}, or -1 is returned if no such value exists.
```
nicktindall, line 786: This is repeated in `@return`. I wonder if we should de-duplicate that and just put a shorter description up here,

src/main/java/net/openhft/chronicle/bytes/Bytes.java
```diff
@@ -780,37 +919,54 @@ default long realCapacity() {
...
+     * at most 1024 bytes
+     */
     @NotNull
     default String toHexString() {
         return toHexString(1024);
     }
 
     /**
+     * Creates and returns a new String representing the contents of this Bytes object in hexadecimal form encoding
+     * at most the provided {@code maxLength } bytes.
```
nicktindall, line 932: Again this description is describing parameters which are described in `@param` I wonder if we should attempt to keep the headline descriptions brief and rely on the other javadoc to tell the detailed story?

nicktindall: (I'm very pro keeping documentation as short and normalised as possible, in the interest of minimising the opportunity for it to go stale)

JerryShea:
Given the potential impact on performance, could we beef up our JMH/JLBH tests first?

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
JerryShea: In the short term: could we change the tests so that they are happy to accept either NPE or IAE? Then we could get phase 1 of this PR in without the changing all the `@NonNull` annotations?

nicktindall (approved):
:+1:

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
nicktindall: I agree the annotation param adds a lot of noise, it would be great to find a way to avoid this.

minborg: Great feedback. The simplest and best solution might be to turn off this feature in IDEA. This means that only real assertions are being made regardless if run under IDEA or Maven and we get consistent behavior and reduced SNR.

We could add a specific test that will trigger unless the feature is turned off and that will output a resolution.

image

src/main/java/net/openhft/chronicle/bytes/Bytes.java
minborg: I could make `@return` much shorter.

src/main/java/net/openhft/chronicle/bytes/Bytes.java
minborg: I think the headline should mention the parameters and how they broadly affect the method and then the `@param` could be more detailed (e.g. non-null, no side effect, stateless...) info. I could try to trim off the headlines a bit and try to move info into the `@param` tags.

minborg:
> Given the potential impact on performance, could we beef up our JMH/JLBH tests first?

I think that would be good but would like to merge this PR ASAP. I will see if I could add some tests where there might be performance regressions. Also, there might be regressions that can only be discovered when used by other libraries.

src/main/java/net/openhft/chronicle/bytes/Bytes.java
minborg: I will raise a separate issue for this

minborg:
In terms of performance, the only thing done was to add a number of `requireNonNull()` and `requireNonNegative()` method calls. These are likely to be inlined by the C2 compiler and would in most cases have zero effect on performance. 

The only thing that I think would be an issue is that we do not have sufficient inline depth (e.g. `MaxInlineLevel`). We would likely not be able to capture those cases in performance tests where CB methods are called directly.  So, I suggest we merge the PR and look for regressions in TC.

These are the methods affected:

```
AppendableUtil
setLength(sb, new Len)

MappedByteStore
write(long offsetInRDO, @NotNull byte[] bytes, int offset, int length)
write(long offsetInRDO, @NotNull ByteBuffer bytes, int offset, int length)
write(long writeOffset, @NotNull RandomDataInput bytes, long readOffset, long length)
write0(long offsetInRDO, @NotNull RandomDataInput bytes, long offset, long length)

NativeBytes
bytesStore(@NotNull BytesStore<Bytes<Underlying>, Underlying> byteStore, long offset, long length)
writeSome(@NotNull final Bytes bytes)

RandomDataInput
read(long offsetInRDI, @NotNull byte[] bytes, int offset, int length)

RandomDataOutput
write(long offsetInRDO, @NotNull byte[] bytes)
write(long offsetInRDO, @NotNull BytesStore bytes)

StreamingDataInput
readWithLength0(long length, @NotNull ThrowingConsumerNonCapturing<S, IORuntimeException, BytesOut> bytesConsumer, StringBuilder sb, BytesOut toBytes)
eadWithLength(long length, @NotNull ThrowingConsumer<S, IORuntimeException> bytesConsumer)
read(@NotNull byte[] bytes, int off, int len)
read(@NotNull ByteBuffer buffer)
read(@NotNull Bytes bytes, int length)
unsafeReadObject(@NotNull Object o, int offset, int length)

StreamingDataOutput
write8bit(@NotNull CharSequence text, int start, int length)
write(@NotNull CharSequence text)
write(@NotNull final CharSequence text,
                    final int startText,
                    final int length)
write(@NotNull RandomDataInput bytes)
write(@NotNull BytesStore bytes)
write(@NotNull byte[] byteArray)

UncheckedBytes
setBytes(@NotNull Bytes bytes)
write(@NotNull RandomDataInput bytes, long offset, long length)
write(@NotNull BytesStore bytes, long offset, long length)
append8bit(@NotNull CharSequence cs)
append8bit(@NotNull char[] chars)
appendUtf8(@NotNull char @NotNull [] chars, int offset, int length)

UncheckedNativeBytes
write(@NotNull RandomDataInput bytes, long offset, long length)
write(long offsetInRDO, byte[] bytes, int offset, int length)
write(long offsetInRDO, @NotNull ByteBuffer bytes, int offset, int length)
write8bit(final @NotNull String text, final int start, final int length)

VanillaBytes
bytesStore(@NotNull final BytesStore<Bytes<Underlying>, Underlying> byteStore,
                               final long offset,
                               final long length)
write(@NotNull RandomDataInput bytes, long offset, long length)
optimisedWrite(@NotNull RandomDataInput bytes, long offset, long length)
write(long position, @NotNull CharSequence str, int offset, int length)
append(@NotNull CharSequence str, int start, int end)
appendUtf8(@NotNull CharSequence str)
append8bit(@NotNull CharSequence cs)
write(@NotNull BytesStore bytes, long offset, long length)
append8bit(@NotNull String cs)
equalBytes(@NotNull BytesStore bytesStore, long length)

ChunkedMappedBytes
 write(final long writeOffset,
                                             @NotNull final RandomDataInput bytes,
                                             long readOffset,
                                             final long length)

write(@NotNull final BytesStore bytes,
                             final long offset,
                             final long length)

appendUtf8(@NotNull char[] chars, int offset, int length)


CommonMappedBytes
write(@NotNull final byte[] bytes,
                                            final int offset,
                                            final int length)
write(@NotNull final RandomDataInput bytes)
write(final long offsetInRDO, @NotNull final RandomDataInput bytes
write(@NotNull final RandomDataInput bytes,
                             final long offset,
                             final long length)
append8bit(@NotNull CharSequence text, int start, int end)
write8bit(@NotNull CharSequence text, int start, int length)
append8bit0(@NotNull String s, int start, int length)
appendUtf8(@NotNull CharSequence cs, int start, int length)

HeapBytesStore
write8bit(long position, @NotNull BytesStore bs)
write8bit(long position, @NotNull String s, int start, int length)
write(long writeOffset,
                                            @NotNull RandomDataInput bytes, long readOffset, long length)
writeLongLength(long writeOffset,
                                 @NotNull RandomDataInput bytes, long readOffset, long length)

NativeBytesStore
 write(
            long offsetInRDO, byte[] bytes, int offset, int length)

write8bit(long position, @NotNull BytesStore bs)
write8bit(long position, @NotNull String s, int start, int length)
copy8bit(byte[] arr, long readPosition, long addressForWrite, long readRemaining)
appendUtf8(long pos, char[] chars, int offset, int length)
read(long offsetInRDI, byte[] bytes, int offset, int length)

SingleMappedBytes
write(final long offsetInRDO,
                                            @NotNull final byte[] bytes,
                                            int offset,
                                            final int length)
write(final long writeOffset,
                                            @NotNull final RandomDataInput bytes,
                                            long readOffset,
                                            final long length)




Bytes
toString(buffer, position, length)
```

PR #259, opened 2021-11-01
Dont create unclosed objects to count, fix sonar badge

nicktindall:
Fixes an issue where Bytes were being created but not closed. Fix the Sonarcloud badge.

### DIFF
--- a/src/test/java/net/openhft/chronicle/bytes/BytesJavaDocComplianceTest.java
+++ b/src/test/java/net/openhft/chronicle/bytes/BytesJavaDocComplianceTest.java
@@ -13,6 +13,7 @@ import java.nio.ByteBuffer;
 import java.nio.ReadOnlyBufferException;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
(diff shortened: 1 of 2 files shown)


src/test/java/net/openhft/chronicle/bytes/BytesJavaDocComplianceTest.java
```diff
@@ -33,15 +34,17 @@
...
             provideBytesObjects()
                     .forEach(args -> {
+                        count.incrementAndGet();
                         final Bytes<Object> bytes = bytes(args);
                         INITIAL_INFO_MAP.put(createCommand(args), new BytesInitialInfo(bytes));
                         releaseAndAssertReleased(bytes);
                         Jvm.pause(100);
                     });
             // Make sure we have unique keys
-            assertEquals(provideBytesObjects().count(), INITIAL_INFO_MAP.size());
```
minborg, line 44: There is was @nicktindall . Well done and thanks ever so much.

minborg (approved):

src/test/java/net/openhft/chronicle/bytes/BytesJavaDocComplianceTest.java
nicktindall: You're very welcome @minborg

PR #261, opened 2021-11-04
Issue/253

minborg:
This PR makes sure all Object::{toString, hashCode, equals} works properly for released Bytes

There is a warning printed if one invokes hashCode/equals that depends on the contents of the Bytes. After some releases, we can change the default behaviour to Object::hashCode and Object::equals. 

Both variants are tested in this PR.


glukos (approved):

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -1258,6 +1257,11 @@ public int byteCheckSum(int start, int end)
         return sum & 0xFF;
     }
 
+    // Only used for testing
```
JerryShea, line 1260: `@TestOnly`

src/test/java/net/openhft/chronicle/bytes/BytesReleaseInvariantObjectTest.java
```diff
@@ -0,0 +1,148 @@
...
+        } else {
+            expected = "";
+        }
+        final String toString = bytes.toString();
+        assertEquals(expected, toString);
+        releaseAndAssertReleased(bytes);
+        assertThrows(ClosedIllegalStateException.class, bytes::toString, createCommand);
+    }
+
+    //@Test
```
JerryShea, line 114: commented-out code

PR #269, opened 2021-11-08
Issue/266

minborg:
This PR adds more than 600 tests and fixes hundreds of issues related to closed resources.


src/main/java/net/openhft/chronicle/bytes/Bytes.java
```diff
@@ -181,7 +182,7 @@
      * Here is an example of field groups:
      * <pre>{@code
      *     static class Padding extends Parent {
-     *         @FieldGroup("p")
+     *         {@literal @}FieldGroup("p")
```
RobAustin, line 185: why the { }

RobAustin (approved):

src/main/java/net/openhft/chronicle/bytes/Bytes.java
minborg: Otherwise JavaDoc will get confused

pom.xml
```diff
@@ -26,7 +26,7 @@
     </parent>
 
     <artifactId>chronicle-bytes</artifactId>
-    <version>2.22ea11-SNAPSHOT</version>
+    <version>2.22ea12-SNAPSHOT</version>
```
JerryShea, line 29: why is this changed?

src/main/java/net/openhft/chronicle/bytes/Bytes.java
minborg: I think the `@FieldGroup`  interferes with the original `@code`

pom.xml
minborg: `develop` moved on and I changed it when reproducing errors in Wire. I should have reverted it back before I created the PR.

PR #270, opened 2021-11-08
Issue/268

minborg:
Expose an internal compare method and add doc and tests


minborg:
It appears this branch was based on another branch rather than `develop`. Sorry.

RobAustin:
image

src/test/java/net/openhft/chronicle/bytes/BytesFactoryUtil.java
```diff
@@ -211,4 +211,65 @@ public void accept(T t) {
     }
 
 
+    static final class BytesInitialInfo {
+        private final long readPosition;
+        private final long writePosition;
+        private final long writeLimit;
+        private final long capacity;
+        private final long realCapacity;
+        //
```
RobAustin: why the //

src/test/java/net/openhft/chronicle/bytes/BytesReleaseInvariantNonPerformantMethodsTest.java
```diff
@@ -0,0 +1,175 @@
...
+        final OutputStream os = new OutputStream() {
+            @Override
+            public void write(int b) throws IOException {
+                throw new UnsupportedEncodingException();
+            }
+        };
+        final BytesStore<?, ?> bs = BytesStore.from(SILLY_NAME);
+        final Bytes<?> bytes = Bytes.from(SILLY_NAME);
+        return Stream.of(
+                NamedConsumer.of(Bytes::toHexString, "toHexString()"),
```
RobAustin: consider using static import for the NamedConsumer

src/test/java/net/openhft/chronicle/bytes/NativeBytesStoreTest.java
```diff
@@ -265,7 +266,7 @@ public void testCopyTo() {
         final Bytes<ByteBuffer> dst = Bytes.elasticByteBuffer();
         try {
             dst.writePosition(src.copyTo(dst));
-            assertEquals(src, dst);
+            assertTrue(BytesInternal.contentEqual(src, dst));
```
RobAustin: suggestion -use AssertEquals(,) but first convert both to a string as this gives you a better diff in the IDE if it fails

pom.xml
```diff
@@ -26,7 +26,7 @@
     </parent>
 
     <artifactId>chronicle-bytes</artifactId>
-    <version>2.22ea11-SNAPSHOT</version>
+    <version>2.22ea12-SNAPSHOT</version>
```
JerryShea: why is this changed? Do we need to rebase?

nicktindall (commented):
Would it be better if we had a pattern like 

```
try (Reference ref = bytes.acquireReferenceOrThrowIfReleased()) {
   ... do stuff
}
```
instead of `throwExceptionIfReleased`, that way we'd be protected against the thing being released while we're in the block as well? I think we discussed something similar in the dev meeting?

It would also have the behaviour that if you were passed a released bytes it would throw.

src/test/java/net/openhft/chronicle/bytes/BytesReleaseInvariantNonPerformantMethodsTest.java
JerryShea: @RobAustin strongly disagree - lots of things have an "of" method - `NamedConsumer.of` reads nicely and is clear

src/test/java/net/openhft/chronicle/bytes/NativeBytesStoreTest.java
minborg: Because we changed how `Bytes::equals` works (it might consider the contents or it might use `==`) then this is the way to go. We might create a custom method (for testing) that asserts content equality and that shows the difference.

minborg: Ahh. I did not read your comment well enough. Good suggestion.

src/test/java/net/openhft/chronicle/bytes/BytesFactoryUtil.java
minborg: No idea... I will remove

minborg:
I have promoted the comment of @nicktindall to an issue: https://github.com/OpenHFT/Chronicle-Bytes/issues/273

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -139,10 +139,16 @@ static boolean contentEqualInt(BytesStore a, BytesStore b)
         }
     }
 
-    public static boolean contentEqual(@Nullable BytesStore a, @Nullable BytesStore b)
-            throws IllegalStateException {
+    public static boolean contentEqual(@Nullable final BytesStore a,
```
RobAustin, line 142: please add java doc

src/main/java/net/openhft/chronicle/bytes/BytesUtil.java
```diff
@@ -151,6 +152,29 @@ public static void writeFile(String file, Bytes<byte[]> bytes)
...
+     * @param a a ByteStore
+     * @param b a ByteStore to be compared with {@code a} for equality
+     * @return if the arguments are equal to each other
+     * @throws ClosedIllegalStateException if either of the provided parameters are released
+     *
+     * @see Object#equals(Object)
+     */
+    public static boolean contentEqual(@Nullable final BytesStore a,
+                                       @Nullable final BytesStore b) {
+        return BytesInternal.contentEqual(a, b);
```
RobAustin, line 175: does this throw IllegalStateException ?  javadoc refers to ClosedIllegalStateException

RobAustin (approved):

src/main/java/net/openhft/chronicle/bytes/BytesUtil.java
minborg: It throws `ClosedIllegalStateException` as specified in the JavaDocs. 

```
    public static void throwExceptionIfReleased(final ReferenceCounted referenceCounted) {
        if (referenceCounted.refCount() <= 0) {
            throw new ClosedIllegalStateException("Released");
        }
    }
```

src/test/java/net/openhft/chronicle/bytes/NativeBytesStoreTest.java
RobAustin: > Ahh. I did not read your comment well enough. Good suggestion.

are you gonna change it again ?

PR #280, opened 2021-11-09
Issue/279

minborg:
This PR reverts back to hashCode/equals always depending on the content of Bytes but adds release checks


minborg:
The build fails because of `acquireReservation() in net.openhft.chronicle.core.io.ReferenceCounted` but this is not merged into `develop` so it is strange that it appears. It just exists in the branch  https://github.com/OpenHFT/Chronicle-Core/commits/issue/273

rogersimmons (approved):

Issue #226, opened 2021-10-12
Bytes.writeSkip should be able to accept -ve numbers


JerryShea:
Customer has reported that MappedBytes.writeSkip used to accept negative numbers but that between bytes 2.20.82 -> 2.21ea61 it has been changed to throw an exception if you give it a negative number.

We call Bytes.writeSkip with a -ve number in a number of places in our codebase so this has to be classed as a bug.

We need to update tests and javadoc for this and readSkip too.

JerryShea:
@minborg this is something to consider as part of your work to improve the method contracts and tests

hft-team-city:
Released in [Chronicle-Bytes-2.21.91](https://github.com/OpenHFT/Chronicle-Bytes/releases/chronicle-bytes-2.21.91), [BOM-2.21.241](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.21.241)

hft-team-city:
Released in [Chronicle-Bytes-2.22ea12](https://github.com/OpenHFT/Chronicle-Bytes/releases/chronicle-bytes-2.22ea12), [BOM-2.22ea57](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.22ea57)

PR #282, opened 2021-11-10
 Bytes.writeSkip should be able to accept -ve numbers #226

JerryShea:


src/test/java/net/openhft/chronicle/bytes/BytesTest.java
```diff
@@ -45,6 +45,7 @@
 import static org.junit.Assert.*;
 import static org.junit.Assume.assumeFalse;
 import static org.junit.Assume.assumeTrue;
+import static org.junit.jupiter.api.Assertions.assertThrows;
```
JerryShea: I did not convert the test to JUnit 5 - just did the smallest thing that could work

nicktindall (approved):

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedBytes.java
```diff
@@ -256,10 +256,10 @@ protected void writeCheckOffset(final long offset, final long adding)
             throws BufferOverflowException, IllegalStateException {
 
         throwExceptionIfClosed();
-        if (offset < 0 || offset > mappedFile.capacity() - adding)
+        if (offset + adding < start() || offset > mappedFile.capacity() - adding)
```
nicktindall, line 259: was this a bug to be using `0` instead of `start()` previously?

JerryShea: Yes I think so

src/test/java/net/openhft/chronicle/bytes/BytesTest.java
```diff
@@ -1049,4 +1049,20 @@ public void testReadWithOffset() {
...
+            String hello = "hello";
+            a.append(hello);
+            assertEquals(hello, a.toString());
+            a.writeSkip(-hello.length());
+            assertEquals("", a.toString());
+            if (!a.unchecked())
+                assertThrows(BufferOverflowException.class, () -> a.writeSkip(-1));
+        } finally {
+            postTest(a);
+        }
```
minborg, line 1066: Nice with a test, preventing any accidental requireNonNegative() to sneak in...

minborg (approved):

PR #285, opened 2021-11-12
Issue/277

minborg:
This PR introduces additional tests and checkings around negative values. This prevents a number of bugs and JVM crashes in the tests.


RobAustin (approved):

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
```diff
@@ -273,12 +277,14 @@ public long capacity() {
     @Override
     public long addressForRead(long offset)
```
nicktindall, line 278: Should we add @NonNegative annotation to these for consistency?

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
```diff
@@ -273,12 +277,14 @@ public long capacity() {
...
             throws UnsupportedOperationException, IllegalStateException, BufferUnderflowException {
+        requireNonNegative(offset);
         return base.addressForRead(offset);
     }
 
     @Override
     public long addressForWrite(long offset)
```
nicktindall, line 285: And here?

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
```diff
@@ -881,6 +896,7 @@ public void readWithLength(Bytes bytes)
     @NotNull
     public Bytes<Void> writePosition(long position)
```
nicktindall, line 897: And here?

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
```diff
@@ -1106,11 +1122,11 @@ private void doIndent() {
 
     @Override
     @NotNull
-    public Bytes<Void> write(@NotNull byte[] bytes, int offset, int length)
+    public Bytes<Void> write(@NotNull byte[] byteArray, int offset, int length)
```
nicktindall, line 1125: Why do we use @NotNull for byte arrays sometimes and not other times (even though we assert not null)

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
```diff
@@ -1655,6 +1674,9 @@ public Bytes write(@NotNull InputStream inputStream)
     @Override
     public Bytes<Void> write(@NotNull BytesStore bytes, long offset, long length)
```
nicktindall, line 1675: @nonnegative

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
```diff
@@ -1709,6 +1731,8 @@ public Bytes write(@NotNull InputStream inputStream)
     @Override
     public void writePositionRemaining(long position, long length)
```
nicktindall, line 1732: @nonnegative

nicktindall (approved):

nicktindall:
Seemed to be some inconsistencies with our use of the annotations, but other than that LGTM!

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
JerryShea: Agree. I believe that offsets should always be from the start of the Bytes and thus always be >= 0

src/main/java/net/openhft/chronicle/bytes/internal/NativeBytesStore.java
```diff
@@ -515,11 +519,15 @@ public long readVolatileLong(long offset) {
...
-    public NativeBytesStore<U> write(
-            long offsetInRDO, byte[] bytes, int offset, int length)
-            throws IllegalStateException {
-        requireNonNull(bytes);
-        memory.copyMemory(bytes, offset, address + translate(offsetInRDO), length);
+    public NativeBytesStore<U> write(@NonNegative final long offsetInRDO,
+                                     final byte[] byteArray,
+                                     @NonNegative final int offset,
+                                     @NonNegative final  int length) throws IllegalStateException {
+        requireNonNegative(offsetInRDO);
```
JerryShea, line 526: My dream is that we (and customers) can validate `@NonNegative` etc at compile time using static analysis, and we can turn off checks at runtime

JerryShea (approved):
I like the additional tests and rigour but I would be much happier if we bolstered our performance tests for common use cases

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
minborg: I think we should. This makes the contract clearer. I have raised https://github.com/OpenHFT/Chronicle-Bytes/issues/290

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
minborg: Originally there was a misunderstanding about what `@NotNull` means for arrays. I raised a new issue about this:
https://github.com/OpenHFT/Chronicle-Bytes/issues/291

src/main/java/net/openhft/chronicle/bytes/internal/NativeBytesStore.java
minborg: This is absolutely doable. We could clone and modify the current non-null plugin.

PR #288, opened 2021-11-15
Issue/287

minborg:
Fixes the build error in TC


src/test/java/net/openhft/chronicle/bytes/BytesReleaseInvariantObjectTest.java
```diff
@@ -49,14 +49,24 @@ void hashcodeContentAffect(final Bytes<?> bytes, final boolean readWrite, final
...
+            if (readWrite) {
+                expected = "The quick brown fox jumped over the usual suspect.";
+                bytes.append(expected);
+            } else {
+                expected = "";
+            }
+            final String toString = bytes.toString();
+            assertEquals(expected, toString);
+        } catch (Exception e) {
+            e.printStackTrace();
```
RobAustin: this can cause too much log output -> e.printStackTrace(); when the tests are run.

RobAustin (approved):

src/test/java/net/openhft/chronicle/bytes/BytesReleaseInvariantObjectTest.java
minborg: Removed

Issue #290, opened 2021-11-16
Consistently introduce @NonNegative across the public API


minborg:

hft-team-city:
Released in [Chronicle-Bytes-2.23ea5](https://github.com/OpenHFT/Chronicle-Bytes/releases/chronicle-bytes-2.23ea5), [BOM-2.23ea27](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.23ea27)

PR #295, opened 2021-11-22
Issue/293 Hoist out boundary checking

minborg:
This PR introduces a concept of packet private `UncheckedRandomDataInput` that, in turn, allows for hoisting out boundary checks from the inner loop. 

Two methods are using this concept in this pr: `BytesInternal::contentEqual` and `BytesInternal::writeFully`. These utility methods are used by several other methods.

A `Bytes` implementation can elect to implement `HasUncheckedRandomData` whereby it will automatically benefit from the optimizations introduced in this PR. To my knowledge, the only implementation _not_ doing so is `HexDumpBytes`.

The proposed API changes are made in anticipation that `AbstractBytes` may be internalized at some future point.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -299,6 +299,18 @@
                     </archive>
                 </configuration>
             </plugin>
+
+
+            <plugin>
+                <groupId>com.github.ferstl</groupId>
+                <artifactId>depgraph-maven-plugin</artifactId>
+                <version>3.3.0</version>
+                <configuration>
+                    <scope>compile</scope>
+                </configuration>
+            </plugin>
+
+
         </plugins>
     </build>
 
--- a/pom.xml
+++ b/pom.xml
@@ -300,17 +300,6 @@
                 </configuration>
             </plugin>
 
-
-            <plugin>
-                <groupId>com.github.ferstl</groupId>
-                <artifactId>depgraph-maven-plugin</artifactId>
-                <version>3.3.0</version>
-                <configuration>
-                    <scope>compile</scope>
-                </configuration>
-            </plugin>
-
-
         </plugins>
     </build>
 
--- a/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
@@ -19,7 +19,7 @@
 package net.openhft.chronicle.bytes;
 
 import net.openhft.chronicle.bytes.internal.BytesInternal;
-import net.openhft.chronicle.bytes.internal.HasUncheckedRandomData;
+import net.openhft.chronicle.bytes.internal.HasUncheckedRandomDataInput;
 import net.openhft.chronicle.bytes.internal.ReferenceCountedUtil;
 import net.openhft.chronicle.bytes.internal.UncheckedRandomDataInput;
 import net.openhft.chronicle.bytes.internal.migration.HashCodeEqualsUtil;
@@ -53,7 +53,7 @@ import static net.openhft.chronicle.core.util.ObjectUtils.requireNonNull;
 @SuppressWarnings("rawtypes")
 public abstract class AbstractBytes<U>
         extends AbstractReferenceCounted
-        implements Bytes<U>, HasUncheckedRandomData {
+        implements Bytes<U>, HasUncheckedRandomDataInput {
     private static final boolean BYTES_BOUNDS_UNCHECKED = Jvm.getBoolean("bytes.bounds.unchecked", false);
 
     // used for debugging
@@ -1274,7 +1274,7 @@ public abstract class AbstractBytes<U>
     }
 
     @Override
-    public UncheckedRandomDataInput acquireUncheckedInput() {
+    public @NotNull UncheckedRandomDataInput acquireUncheckedInput() {
         return uncheckedRandomDataInput;
     }
 
--- a/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
@@ -135,10 +135,11 @@ public abstract class AbstractBytes<U>
     @Override
     public Bytes<U> clearAndPad(long length)
             throws BufferOverflowException {
-        if ((start() + length) > capacity()) {
-            throw newBOERange(start(), length, "clearAndPad failed. Start: %d + length: %d > capacity: %d", capacity());
+        final long start = start();
+        if ((start + length) > capacity()) {
+            throw newBOERange(start, length, "clearAndPad failed. Start: %d + length: %d > capacity: %d", capacity());
         }
-        long l = start() + length;
+        long l = start + length;
         readPosition = l;
         uncheckedWritePosition(l);
         writeLimit = capacity();
@@ -784,7 +785,8 @@ public abstract class AbstractBytes<U>
 
     private void writeCheckOffset0(long offset, long adding)
             throws DecoratedBufferOverflowException {
-        if (offset < start() || offset + adding < start()) {
+        final long start = start();
+        if (offset < start || offset + adding < start) {
             throw newBOELower(offset);
         }
         if ((offset + adding) > writeLimit()) {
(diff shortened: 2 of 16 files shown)


pom.xml
```diff
@@ -299,6 +299,18 @@
                     </archive>
                 </configuration>
             </plugin>
+
+
+            <plugin>
+                <groupId>com.github.ferstl</groupId>
+                <artifactId>depgraph-maven-plugin</artifactId>
+                <version>3.3.0</version>
```
RobAustin: can we move this to the third party BOM

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -1174,7 +1179,7 @@ public int hashCode() {
 
     @Override
     public boolean equals(Object obj) {
-        return HashCodeEqualsUtil.contentEquals(this, obj);
+        return obj instanceof BytesStore && BytesInternal.contentEqual(this, (BytesStore) obj);
```
RobAustin, line 1182: maybe static import for "BytesInternal.contentEqual"

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -101,19 +102,36 @@
...
+            // The size is different so, we know that a and b cannot be equal
+            return false;
+        return readRemaining <= Integer.MAX_VALUE
+                ? contentEqualInt(a, b)
+                : contentEqualsLong(a, b);
+    }
+
     /**
-     * Optimise for the common case where the length is 31-bit.
+     * a >= b here and we also know it is safe to read bLength
```
RobAustin: is this valid java doc, is it not missing its @params

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -140,32 +158,99 @@ static boolean contentEqualInt(BytesStore a, BytesStore b)
...
-    public static boolean contentEqual(@Nullable final BytesStore a,
-                                       @Nullable final BytesStore b) throws IllegalStateException {
-        if (a == null) return b == null;
-        if (b == null) {
-            // The contract stipulates that if either ByteStores are closed then we throw an Exception
-            throwExceptionIfReleased(a);
-            return false;
+    /**
+     * Optimise for the common case where the length is 31-bit.
+     */
```
RobAustin: add javadoc for @params

src/main/java/net/openhft/chronicle/bytes/internal/HasUncheckedRandomData.java
```diff
@@ -0,0 +1,7 @@
+package net.openhft.chronicle.bytes.internal;
+
```
RobAustin: can this be 

```
@Functional
```

src/main/java/net/openhft/chronicle/bytes/internal/HeapBytesStore.java
```diff
@@ -482,9 +481,9 @@ public long write8bit(long position, @NotNull String s, int start, int length) {
     @NotNull
     @Override
     public HeapBytesStore<U> write(@NonNegative final long offsetInRDO,
-                                    final byte[] byteArray,
-                                    @NonNegative final int offset,
-                                    @NonNegative final int length) throws BufferOverflowException {
+                                   final byte[] byteArray,
```
RobAustin, line 484: can this be @NotNull

src/main/java/net/openhft/chronicle/bytes/internal/NativeBytesStore.java
```diff
@@ -705,7 +704,7 @@ public long readIncompleteLong(long offset) {
 
     @Override
     public boolean equals(Object obj) {
-        return HashCodeEqualsUtil.contentEquals(this, obj);
+        return obj instanceof BytesStore && BytesInternal.contentEqual(this, (BytesStore) obj);
     }
 
     public void setAddress(long address) {
```
RobAustin, line 710: maybe add a @NotNegitive ( or something like that ! )

src/main/java/net/openhft/chronicle/bytes/internal/UncheckedRandomDataInput.java
```diff
@@ -0,0 +1,59 @@
...
+     * <p>
+     * Memory bounds must be checked before invoking this method or else the result
+     * is undefined and may lead to JVM crashes.
+     *
+     * @param offset to read
+     * @return the long
+     * @throws BufferUnderflowException if the offset is outside the limits of the Bytes
+     * @throws IllegalStateException    if released
+     */
+    byte readByte(long offset);
```
RobAustin, line 18: maybe add a @NotNegitive ( or something like that ! ) for all the methods below

RobAustin (approved):

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -140,32 +158,99 @@ static boolean contentEqualInt(BytesStore a, BytesStore b)
...
-    private static boolean contentEqualsLong(@NotNull BytesStore a, @NotNull BytesStore b) {
+    /**
+     * a >= b here and we also know it is safe to read bLength
+     */
+    static <U extends BytesStore<?, ?> & HasUncheckedRandomData>
+    boolean contentEqualIntUnchecked(@NotNull final U a,
+                                     @NotNull final U b,
+                                     @NonNegative final int aLength,
+                                     @NonNegative final int bLength) throws IllegalStateException {
+
```
nicktindall: Should we add an assertion that aLength >= bLength here? just in case?

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -140,32 +158,99 @@ static boolean contentEqualInt(BytesStore a, BytesStore b)
...
+    }
+
+    /**
+     * a >= b here and we also know it is safe to read bLength
+     */
+    private static boolean contentEqualsLong(@NotNull final BytesStore a,
+                                             @NotNull final BytesStore b,
+                                             @NonNegative final long aLength,
+                                             @NonNegative final long bLength) {
         // assume a >= b
```
nicktindall, line 250: and here?

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -185,7 +270,46 @@ private static boolean contentEqualsLong(@NotNull BytesStore a, @NotNull BytesSt
...
+
+    /**
+     * a >= b here and we also know it is safe to read bLength
+     */
+    private static <U extends BytesStore<?, ?> & HasUncheckedRandomData>
+    boolean contentEqualsLongUnchecked(@NotNull final U a,
+                                       @NotNull final U b,
+                                       @NonNegative final long aLength,
+                                       @NonNegative final long bLength) {
+
```
nicktindall, line 286: and here?

nicktindall (approved):

pom.xml
minborg: It is added to the root-pom now.

src/main/java/net/openhft/chronicle/bytes/internal/NativeBytesStore.java
minborg: There is a separate issue for this: https://github.com/OpenHFT/Chronicle-Bytes/issues/290

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -140,32 +158,99 @@ static boolean contentEqualInt(BytesStore a, BytesStore b)
...
-            throwExceptionIfReleased(a);
-            return false;
+    /**
+     * Optimise for the common case where the length is 31-bit.
+     */
+    static <U extends BytesStore<?, ?> & HasUncheckedRandomData>
+    boolean contentEqualInt(@NotNull final BytesStore<?, ?> a,
+                            @NotNull final BytesStore<?, ?> b) throws IllegalStateException {
+
+        final int aLength = (int) a.realReadRemaining();
```
JerryShea: shouldn't we assert that aLength and bLength are actually in int range?

src/main/java/net/openhft/chronicle/bytes/internal/NativeBytesStore.java
JerryShea: for address? addresses can be -ve

minborg: True. Addresses can be negative (e.g. for mapped files in Linux) but offsets and lengths are always non-negative.

PR #301, opened 2021-11-23
Take readPosition into account when copying to readLimit

### LINKED ISSUE #300: Bytes.copy, copyTo should take readPosition into account
Right now, both will copy `readRemaining()` bytes, but will start from position 0 instead of `readPosition()`, which does not make sense:

```
    @Test
    public void copyShouldCopy() {
 
        final Bytes<byte[]> src = Bytes.from("hello again");
        src.readSkip(7);
        assertThat(src.copy()).isEqualTo(src);
    }
 
    @Test
    public void copyToShouldStartFromTheReadPositionIfItUsesReadRemaining() {
 
        final Bytes<byte[]> src = Bytes.from("hello again");
        src.readSkip(7);
 
        final byte[] buffer = new byte[100];
        final int copiedLen = src.copyTo(buffer);
        assertThat(new String(buffer, 0, copiedLen)).isEqualTo(src.toString());
    }
```

alamar:
fixes #300

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/RandomDataInput.java
+++ b/src/main/java/net/openhft/chronicle/bytes/RandomDataInput.java
@@ -301,7 +301,7 @@ public interface RandomDataInput extends RandomCommon {
             throws BufferUnderflowException, IllegalStateException;
 
     /**
-     * Read a byte[] from memory.
+     * Read a byte[] from memory from {@link #readPosition()} to {@link #readLimit()}.
      *
      * @return the length actually read.
      * @throws BufferUnderflowException if the offset is outside the limits of the Bytes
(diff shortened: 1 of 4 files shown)


src/main/java/net/openhft/chronicle/bytes/RandomDataInput.java
```diff
@@ -313,7 +313,7 @@ default int copyTo(@NotNull byte[] bytes)
         throwExceptionIfReleased(this);
         int len = (int) Math.min(bytes.length, readRemaining());
         for (int i = 0; i < len; i++)
-            bytes[i] = readByte(start() + i);
+            bytes[i] = readByte(readPosition() + i);
```
minborg: We could move out readPosition() from the inner loop.

minborg:
These changes look right but may affect the existing code.

minborg (approved):

src/test/java/net/openhft/chronicle/bytes/BytesInternalTest.java
```diff
@@ -327,6 +326,24 @@ public void testParseDouble() {
         }
     }
 
+    @Test
+    public void testCopyAfterSkip() {
+        final Bytes<byte[]> src = Bytes.from("hello again");
+
+        src.readSkip(7);
```
JerryShea, line 333: I don't understand. Doesn't readSkip increase the read position? Shouldn't `src.copy()` be "gain"?

alamar: I think so! But it used to be "hell" before fix.

PR #302, opened 2021-11-23
Issue/299

minborg:
This PR reduces the probability of unrelated synchronization. See original issue

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,7 +151,7 @@
                         <configuration>
                             <referenceVersion>2.22ea3</referenceVersion>
                             <artifactsURI>https://teamcity.chronicle.software/repository/download</artifactsURI>
-                            <binaryCompatibilityPercentageRequired>83.6</binaryCompatibilityPercentageRequired>
+                            <binaryCompatibilityPercentageRequired>81.3</binaryCompatibilityPercentageRequired>
                         </configuration>
                     </execution>
                 </executions>
--- a/src/main/java/net/openhft/chronicle/bytes/MappedFile.java
+++ b/src/main/java/net/openhft/chronicle/bytes/MappedFile.java
@@ -47,7 +47,8 @@ import java.nio.channels.FileLock;
 public abstract class MappedFile extends AbstractCloseableReferenceCounted {
     static final boolean RETAIN = Jvm.getBoolean("mappedFile.retain");
     private static final long DEFAULT_CAPACITY = 128L << 40;
-    protected final String internalizedToken;
+    // The canonical path is pre-pended with static and random data to prevent unrelated synchronization on internalized Strings
+    protected final String canonicalPath;
     @NotNull
     private final File file;
     private final boolean readOnly;
@@ -57,7 +58,7 @@ public abstract class MappedFile extends AbstractCloseableReferenceCounted {
                          final boolean readOnly)
             throws IORuntimeException {
         this.file = file;
-        this.internalizedToken = CanonicalPathUtil.of(file);
+        this.canonicalPath = CanonicalPathUtil.of(file);
         this.readOnly = readOnly;
     }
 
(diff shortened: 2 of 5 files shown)


src/main/java/net/openhft/chronicle/bytes/MappedFile.java
```diff
@@ -46,6 +47,7 @@
 public abstract class MappedFile extends AbstractCloseableReferenceCounted {
     static final boolean RETAIN = Jvm.getBoolean("mappedFile.retain");
     private static final long DEFAULT_CAPACITY = 128L << 40;
+    // The canonical path is pre-pended with static and random data to prevent unrelated synchronization on internalized Strings
     protected final String canonicalPath;
```
alamar: Maybe it should be called pathLock? It's not path anymore.

minborg: Yepp. I thought that too but then the compatibility plugin complained as this variable is in the API. We can revisit once we have internalized it.

alamar: In this case it is better to rename it, if any users of this class happen to use this variable as a path, they should definitely rewrite their code now, since it is no good anymore. It has protected access so renaming will no longer be purely decorative.

Compatibility plugin deliver value but has to be tuned down.

alamar (changes requested):

src/main/java/net/openhft/chronicle/bytes/MappedFile.java
alamar: After it is renamed to a descriptive name the comment may be shortened or dropped.

minborg: This is fixed now

pom.xml
```diff
@@ -151,7 +151,7 @@
                         <configuration>
                             <referenceVersion>2.22ea3</referenceVersion>
                             <artifactsURI>https://teamcity.chronicle.software/repository/download</artifactsURI>
-                            <binaryCompatibilityPercentageRequired>83.6</binaryCompatibilityPercentageRequired>
+                            <binaryCompatibilityPercentageRequired>81.3</binaryCompatibilityPercentageRequired>
```
RobAustin, line 154: doh - not good :-(

src/main/java/net/openhft/chronicle/bytes/MappedFile.java
```diff
@@ -46,7 +47,8 @@
 public abstract class MappedFile extends AbstractCloseableReferenceCounted {
     static final boolean RETAIN = Jvm.getBoolean("mappedFile.retain");
     private static final long DEFAULT_CAPACITY = 128L << 40;
-    protected final String canonicalPath;
+    // The canonical path is pre-pended with static and random data to prevent unrelated synchronization on internalized Strings
+    protected final String internalizedToken;
```
RobAustin: I understand why you changed this, - so, If you think this should be private, please make it private and provided access to depending class via a method call, instead of having a protected field.

RobAustin (changes requested):
please make "internalizedToken" private and wrap its access via a method

pom.xml
alamar: As explained in comments, that's better than getting runtime errors for our users. Maybe we could bump it to x.23 if it's planned any time soon.

alamar (approved):

src/main/java/net/openhft/chronicle/bytes/internal/CanonicalPathUtil.java
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * Returns an internalized String that is unlikely to collide with other internalized strings in the JVM.
+     *
+     * @param file to use
+     * @return internalized String
+     */
+    public static String of(@NotNull final File file) {
+        try {
+            return (SUFFIX + "::" + file.getCanonicalPath())
+                    .intern();
```
nicktindall, line 26: bike-shedding, but shouldn't this be `PREFIX` if it's being pre-pended?

nicktindall (approved):

src/main/java/net/openhft/chronicle/bytes/internal/CanonicalPathUtil.java
minborg: You are perfectly right

RobAustin (approved):

PR #303, opened 2021-11-23
Issue/298

minborg:
This PR removes most of the issues reported as "Bugs" in SQ. There is a slight behaviour change in the unlikely event we have memory regions closed from outside within the method itself whereby an Exeption rather than an Error is thrown. In most (if not all) cases this can never happen.

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/Bytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/Bytes.java
@@ -62,9 +62,8 @@ import static net.openhft.chronicle.core.util.ObjectUtils.requireNonNull;
  * Generally, a Bytes object is a resource that is ReferenceCounted and certain operations invoked on a Bytes
  * object that has been released, might result in an {@link IllegalStateException} being thrown.
  *
- * @see BytesStore
  * @param <U> Underlying type
- *
+ * @see BytesStore
  */
 @SuppressWarnings({"rawtypes", "unchecked"})
 public interface Bytes<U> extends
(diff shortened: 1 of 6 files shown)


RobAustin (approved):

src/main/java/net/openhft/chronicle/bytes/Bytes.java
```diff
@@ -62,9 +62,8 @@
  * Generally, a Bytes object is a resource that is ReferenceCounted and certain operations invoked on a Bytes
  * object that has been released, might result in an {@link IllegalStateException} being thrown.
  *
- * @see BytesStore
  * @param <U> Underlying type
- *
+ * @see BytesStore
```
alamar, line 66: @see does usually go before @param's

src/main/java/net/openhft/chronicle/bytes/internal/AbstractBytesStore.java
```diff
@@ -53,6 +53,12 @@ public int hashCode() {
         return BytesStoreHash.hash32(this);
     }
 
+    @SuppressWarnings("rawtypes")
+    @Override
+    public boolean equals(Object obj) {
```
alamar, line 58: Why do we resurrect it here, and what's with hashCode()?

src/main/java/net/openhft/chronicle/bytes/Bytes.java
minborg: I think the order as it is now complies with https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#orderoftags

src/main/java/net/openhft/chronicle/bytes/internal/AbstractBytesStore.java
minborg: Classes that overrides hashCode() should also override equals(). Hence, I have moved the declaration so they appear in pairs.

PR #305, opened 2021-11-23
Add assertions to test, Fix #304

### LINKED ISSUE #304: Add assertions to tests

minborg:


src/test/java/net/openhft/chronicle/bytes/Bytes4Test.java
```diff
@@ -22,7 +22,11 @@ public void segFault() {
         BytesStore bs = BytesStore.from("this is a test          ");
         bs.append(14, 16, 12);
         bs.isClear();
-        bs.writeUtf8(14, "this is a another text it should over write the other");
+        try {
+            bs.writeUtf8(14, "this is a another text it should over write the other");
+        } catch (Throwable throwable) {
+            fail(throwable.getMessage());
```
RobAustin: why do this, if the test throws an exception it will fail the test anyway

src/test/java/net/openhft/chronicle/bytes/BytesInternalTest.java
```diff
@@ -453,7 +453,16 @@ public void testNoneDirectWritePerformance() {
...
+
 */
+            // This is a performance test so just assert it ran
+            assertTrue(time1 > 0);
+            assertTrue(time2 > 0);
+            assertTrue(time3 > 0);
+            assertTrue(time4 > 0);
+            assertTrue(time5 > 0);
+            assertTrue(time6 > 0);
             Thread.yield();
```
RobAustin, line 466: Why .yield() !

src/test/java/net/openhft/chronicle/bytes/MappedFileTest.java
```diff
@@ -42,7 +42,11 @@
 
     @Test
     public void testWarmup() {
-        MappedFile.warmup();
+        try {
+            MappedFile.warmup();
+        } catch (Throwable t) {
+            fail(t.getMessage());
```
RobAustin: Suggestion :  rewrite using JUnit5, ( something like assertNotThrows )  as this is just adding code clutter.

RobAustin (changes requested):
I suggest we remove the fail(..) blocks as this just add code clutter and JUnit5 may have better options.

nicktindall (approved):
Agree with Rob on the explicit fails

src/test/java/net/openhft/chronicle/bytes/MappedBytesTest.java
```diff
@@ -484,6 +484,12 @@ public void multiBytes()
             original.writeInt(54, 12345678);
 //            System.out.println("Original(54): " + original.readInt(54));
```
JerryShea, line 485: can we get rid of commented-out code?

src/test/java/net/openhft/chronicle/bytes/BytesInternalTest.java
minborg: No idea...

src/test/java/net/openhft/chronicle/bytes/MappedMemoryTest.java
```diff
@@ -134,8 +135,11 @@ public void withRawNativeBytesTess()
                     }
                     bytes.releaseLast(test);
 
+                    Jvm.perf().on(getClass(), "With NativeBytes,\t\t time= " + 80 * (System.nanoTime() - startTime) / BLOCK_SIZE / 10.0 + " ns, number of longs written=" + BLOCK_SIZE / 8);
+                } catch (Throwable throwable) {
+                    // Performance test so just make sure the test ran
+                    fail(throwable.getMessage());
```
RobAustin, line 141: why not use assertDoesNotThrow here was well ?

RobAustin (approved):

src/test/java/net/openhft/chronicle/bytes/Bytes4Test.java
JerryShea: yes

src/test/java/net/openhft/chronicle/bytes/MappedMemoryTest.java
minborg: It will take some time to fix so I left it as is.

PR #315, opened 2021-12-10
Replace regexp with state machine, Fix #306

### LINKED ISSUE #306: Rewrite PropertyReplacer to avoid catastrophic backtracking
See:

https://sonarcloud.io/project/security_hotspots?id=OpenHFT_Chronicle-Bytes#

minborg:
Most of the regular expression engines use backtracking to try all possible execution paths of the regular expression when evaluating an input, in some cases, it can cause performance issues, called *catastrophic backtracking* situations. In the worst case, the complexity of the regular expression is exponential in the size of the input, this means that a small carefully-crafted input (like 20 chars) can trigger catastrophic backtracking and cause a denial of service of the application. Super-linear regex complexity can lead to the same impact too with, in this case, a large carefully-crafted input (thousands of chars).

This PR replaces the use of regexp with a custom-made state machine, thereby preventing catastrophic backtracking.

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/util/PropertyReplacer.java
+++ b/src/main/java/net/openhft/chronicle/bytes/util/PropertyReplacer.java
@@ -18,67 +18,190 @@
 package net.openhft.chronicle.bytes.util;
 
 import net.openhft.chronicle.bytes.BytesUtil;
+import net.openhft.chronicle.core.annotation.NonNegative;
 import org.jetbrains.annotations.NotNull;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Properties;
-import java.util.regex.Matcher;
+import java.util.Scanner;
 import java.util.regex.Pattern;
+
+import static net.openhft.chronicle.core.util.Ints.requireNonNegative;
+import static net.openhft.chronicle.core.util.ObjectUtils.requireNonNull;
 
 public enum PropertyReplacer {
     ; // none
 
-    private static final Pattern EXPRESSION_PATTERN = Pattern.compile("\\$\\{\\s*([^}]*?)\\s*\\}");
-
-    public static String replaceTokensWithProperties(String expression) throws IllegalArgumentException {
-
-        StringBuilder result = new StringBuilder(expression.length());
-        int i = 0;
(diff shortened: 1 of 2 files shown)


JerryShea (changes requested):
PropertyReplacer is not performance critical code. I would prefer we left this code as is

minborg:
The reason for this PR is not performance issues. Whether or not the procedure is 1-5 times faster or slower is not relevant. What is relevant is that crafted inputs (e.g. received over the wire or from a local file) can completely halt the application virtually forever. 

That is the reason for the suggested changes. 

Read more here: https://sonarcloud.io/organizations/openhft/rules?open=java%3AS5852&rule_key=java%3AS5852

JerryShea:
I just don't buy the cost/benefit of adding hundreds of lines of code to mitigate a problem which can happen when parsing configuration files. Regexps have their limitations but are an appropriate solution for this problem.

nicktindall:
I agree with Jerry on this, the linked vulnerability is an OWASP vulnerability, the "WA" being web application, if we were using this code to process inputs from the outside world in a process that absolutely had to remain live (like in a web server), I could see it being worth the added complexity, but if I understand it correctly this code only processes config files written by the in-house teams at startup, so:
1. It's unlikely they'd intentionally try to craft a bad input
2. If they did it would just prevent the app starting up, not catastrophically bring down a running mission critical app

In saying that, if we were doing a hot reload of config in a running process and the in-house team (or config file generator) accidentally produced a bad file (I'm not sure how feasible this is, it sounds like you'd need to do something pretty pathological to trigger this), it could be an issue.

Did we look at modifying the regex to reduce the possibility of back-tracking? (e.g. by putting a reasonable limit on the variable length or such?)

RobAustin:
@minborg can we have a chat to discuss the cost benefits of these changes? I feel we may be approaching the threshold where it's not worth making any further changes as the benefits are negligible.

minborg:
I agree with your arguments above. I have created a new PR with another way of solving the problem. See https://github.com/OpenHFT/Chronicle-Bytes/pull/320

peter-lawrey:
I think we should be reducing the complexity of the code, can we turn that rule off.

minborg:
This suggestion was dropped in favour of providing an updated regexp.

PR #317, opened 2021-12-10
Remove issues flagged as Security Hotspots. Fix #316

### LINKED ISSUE #316: Remove issues flagged as Security Hotspots
- [ ] `NativeBytes::ensureCapacity` invokes `Exception::printStackTrace`

- [ ] `ChunkedMappedFile` is using a shared tmp resource

minborg:


src/main/java/net/openhft/chronicle/bytes/NativeBytes.java
```diff
@@ -194,7 +194,6 @@ public void ensureCapacity(final long desiredCapacity)
             writeCheckOffset(desiredCapacity, 0);
         } catch (BufferOverflowException e) {
             IllegalArgumentException iae = new IllegalArgumentException("Bytes cannot be resized to " + desiredCapacity + " limit: " + capacity(), e);
-            iae.printStackTrace();
```
JerryShea, line 197: this is there for a reason - let's just suppress the warning

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedFile.java
```diff
@@ -83,7 +83,7 @@ public static void warmup() {
         try {
             Jvm.setExceptionHandlers(Slf4jExceptionHandler.ERROR, null, null);
 
-            @NotNull final File file = File.createTempFile("delete_warming_up", "me");
+            final File file = File.createTempFile("delete_warming_up", "me", new File("/scratch_area"));
```
JerryShea: but will this actually warm up the code in the same way?

src/main/java/net/openhft/chronicle/bytes/NativeBytes.java
minborg: I suggest using a real logger. I will update accordingly.

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedFile.java
minborg: Yes, I think it will. Merely placing the queue files one level down in the directory structure will likely not change the warmup outcome at all.  I will, however, improve on how the temp dir is created in a new proposal.

nicktindall (approved):

PR #320, opened 2021-12-14
Rewrite regexp. Fix #319

### LINKED ISSUE #319: Replace regexp preventing backtracing

minborg:
This PR fixed the issue addressed in https://github.com/OpenHFT/Chronicle-Bytes/pull/315 by using a possessive construct with mutually excluded paths in the regexp, thereby preventing backtracing. 

This PR, if accepted, would remove the need of merging https://github.com/OpenHFT/Chronicle-Bytes/pull/315 which could be closed instead.

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/util/PropertyReplacer.java
+++ b/src/main/java/net/openhft/chronicle/bytes/util/PropertyReplacer.java
@@ -29,7 +29,7 @@ import java.util.regex.Pattern;
 public enum PropertyReplacer {
     ; // none
 
-    private static final Pattern EXPRESSION_PATTERN = Pattern.compile("\\$\\{\\s*([^}]*?)\\s*\\}");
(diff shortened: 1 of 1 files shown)


nicktindall (approved):
Very nice, but does putting the \\s inside the square brackets mean we get spaces in identifiers?

JerryShea (approved):

minborg:
> Very nice, but does putting the \s inside the square brackets mean we get spaces in identifiers?

No. It means that all characters except '}' and white spaces are consumed by this group.

nicktindall:
> > Very nice, but does putting the \s inside the square brackets mean we get spaces in identifiers?
> 
> No. It means that all characters except '}' and white spaces are consumed by this group.

Ah I see, the caret applies to the whole contents of the square brackets, not just the next symbol.

PR #321, opened 2021-12-22
Added support for using a LongValue as a timed lock which can check if a thread is running

peter-lawrey:
https://github.com/OpenHFT/Chronicle-Core/issues/301


minborg:
it does not build?

peter-lawrey:
It didn't work as it was waiting on a pull request in Core

PR #325, opened 2022-01-05
Implement subSequence(start, end) in Bytes

### LINKED ISSUE #324: Implement subSequence(start, end) in Bytes
```
    /**
     * Not supported.
     */
    @NotNull
    @Override
    default CharSequence subSequence(int start, int end) {
        throw new UnsupportedOperationException("todo");
    }
```

glukos:


src/test/java/net/openhft/chronicle/bytes/BytesStoreTest.java
```diff
@@ -23,4 +23,57 @@ public void from2() {
...
+    }
+
+    private void testSubSequence(Bytes<?> hello) {
+        CharSequence helloSubsequence = hello.subSequence(0, 5);
+        assertEquals("Hello", helloSubsequence.toString());
+
+        CharSequence hellSubsequence = hello.subSequence(0, 4);
+        assertEquals("Hell", hellSubsequence.toString());
+
+        CharSequence ellSubsequence = hello.subSequence(1, 4);
```
minborg, line 54: Brilliant variable name...

glukos: Didn't want to steal fame - this is Peter's idea from another test method :grinning:

src/main/java/net/openhft/chronicle/bytes/BytesStore.java
```diff
@@ -397,12 +397,19 @@ default char charAt(int index)
     }
 
     /**
-     * Not supported.
+     * This method is inherited from CharSequence.
```
minborg: I think we should not add a JavaDoc here as it hides the original one from CharSequence. Maybe put it as a comment instead.

src/main/java/net/openhft/chronicle/bytes/BytesStore.java
```diff
@@ -397,12 +397,19 @@ default char charAt(int index)
...
+     *
+     * @return  {@link BytesStore} of (end - start) length
+     * containing chars from #charAt(start) to #charAt(end - 1), inclusive.
+     * Memory storage type (heap or native) is preserved.
      */
     @NotNull
     @Override
     default CharSequence subSequence(int start, int end) {
```
minborg: We could return a more specific CharSequence namely ByteStore

glukos: Agreed, changed return type to BytesStore (Bytes would be incorrect)

src/main/java/net/openhft/chronicle/bytes/BytesStore.java
```diff
@@ -397,12 +397,19 @@ default char charAt(int index)
...
-        throw new UnsupportedOperationException("todo");
+        if (start < 0 || end > length() || end < start)
+            throw new IllegalArgumentException("start " + start + ", end " + end + ", length " + length());
+
+        return subBytes(readPosition() + start, end - start);
```
minborg: use (long)end to perform operations using `long` rather than `int`

src/main/java/net/openhft/chronicle/bytes/BytesStore.java
minborg: I think we should adhere to the contract of subSequence and throw an `IndexOutOfBoundsException` if start or end are negative, if end is greater than length(), or if start is greater than end

src/main/java/net/openhft/chronicle/bytes/BytesStore.java
glukos: Personally I like 
```
/**
 * {@inheritDoc}
 */
```
in cases like this.

However, in this particular case I find the comment useful. At least, if there were only inherited doc, I'd check which BytesStore type is constructed.

src/main/java/net/openhft/chronicle/bytes/BytesStore.java
glukos: We know in advance that `end >= start`. Is there another reason to cast except for overflow protection?

minborg (approved):

PR #328, opened 2022-01-11
Add lenient checking to readByte()

alamar:


minborg (approved):

JerryShea (approved):
looks reasonable!

alamar:
Asking for re-review since I also had to rewrite the Double parser, which was not using limit checks consistently. It is debatable which one of read byte methods it should really use.

Please review @minborg 

Is it normal that it would consume a first non-digit char, though? It is expected that it is a newline or comma, but still.

minborg (approved):

minborg:
We are looking at the first char two times so it might be marginally slower?

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -406,7 +406,10 @@ public byte readByte()
             return bytesStore.readByte(offset);
 
         } catch (BufferUnderflowException e) {
-            return 0;
+            if (lenient) {
+                return 0;
```
minborg, line 410: This would be superslow as an exception is created on each read...

alamar: Sure, but the parseDouble() no longer uses this method, it is the one for end-users of Bytes and in line with other methods available for them.

alamar:
@JerryShea can you please re-review?

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -2522,36 +2523,39 @@ public static double parseDouble(@NotNull StreamingDataInput in)
...
-                    negative = true;
-                    ch = in.rawReadByte();
                     break;
                 default:
                     // do nothing
             }
             int tens = 0;
-            while (true) {
+            while (in.readRemaining() > 0) {
+                ch = in.readUnsignedByte() & 0xFF;
```
JerryShea, line 2558: do we have a performance test for parseDouble method? If not we definitely should have

src/test/java/net/openhft/chronicle/bytes/readme/PrimitiveTest.java
```diff
@@ -214,10 +214,13 @@ public void testTextPrimitive() {
             bytes.append('3').append('\n');
             bytes.append(4.1f).append('\n');
             bytes.append(5.2).append('\n');
+            bytes.append(Double.NEGATIVE_INFINITY).append('\n');
```
JerryShea, line 217: cool

JerryShea (approved):
Like the changes very much - just would be good to see evidence that parseDouble performance has not regressed :)

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
alamar: Yes, we do (AppendLongCoolerMain) and it does not show any serious difference:

Before
> PAUSE1,band,<0.1,<1,<10,<100, 1.6, 99.9, 99.9, 100.0,%iles,50/90 97/99 99.7/99.9 99.97/99.99 - worst was 0.226 / 0.375  0.451 / 0.489  0.538 / 0.702  8.67 / 12.32 - 21.2
> BUSY100,band,<0.1,<1,<10,<100, 83.5, 99.8, 100.0, 100.0,%iles,50/90 97/99 99.7/99.9 99.97/99.99 - worst was 0.066 / 0.117  0.174 / 0.355  0.598 / 1.89  1.89 / 1.89 - 1.89
> ALL,band,<0.1,<1,<10,<100, 0.0, 99.7, 100.0, 100.0,%iles,50/90 97/99 99.7/99.9 99.97/99.99 - worst was 0.373 / 0.598  0.690 / 0.822  0.918 / 9.63  9.76 / 9.76 - 9.76

After
> PAUSE1,band,<0.1,<1,<10,<100, 1.7, 99.9, 99.9, 100.0,%iles,50/90 97/99 99.7/99.9 99.97/99.99 - worst was 0.212 / 0.365  0.431 / 0.481  0.534 / 0.646  8.29 / 9.06 - 21.7
> BUSY100,band,<0.1,<1,<10,<100, 83.1, 100.0, 100.0, 100.0,%iles,50/90 97/99 99.7/99.9 99.97/99.99 - worst was 0.065 / 0.129  0.234 / 0.391  0.546 / 0.562  0.562 / 0.562 - 0.562
> ALL,band,<0.1,<1,<10,<100, 0.0, 99.4, 100.0, 100.0,%iles,50/90 97/99 99.7/99.9 99.97/99.99 - worst was 0.465 / 0.714  0.826 / 0.942  1.060 / 1.132  1.188 / 1.188 - 1.188

JerryShea: love it @alamar

JerryShea (approved):
Thought I had already approved but approving again. So good I approved it twice!

PR #331, opened 2022-02-14
Use compatibility enforcer from parent, fix warnings in tests

alamar:


JerryShea (approved):

src/test/java/net/openhft/chronicle/bytes/ref/ByteableReferenceTest.java
```diff
@@ -54,7 +56,7 @@ public void shouldMakeReservationOnCurrentStore() {
         try {
             firstStore.writeLong(0, 17);
             final BytesStore secondStore = BytesStore.nativeStore(64);
-            try {
+            try (AbstractReference byteable = byteableCtor.get()) {
```
minborg, line 59: Nice!

minborg (approved):

PR #333, opened 2022-02-18
Add lenient checking to readByte()

### LINKED ISSUE #329: Add lenient checking to readByte()
Right now, `readShort()` to `readDouble()` have lenient checking in `AbstractBytes`, but `readByte()` does not and will always return 0.

Let's make it aware of `lenient` setting. The rest of unchecked/unsigned byte methods which return -1 may remain the same.

alamar:
change double parsing routine to enforce limit checks

closes #329

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
@@ -406,7 +406,10 @@ public abstract class AbstractBytes<U>
             return bytesStore.readByte(offset);
 
         } catch (BufferUnderflowException e) {
-            return 0;
+            if (lenient) {
+                return 0;
+            }
+            throw e;
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -406,7 +406,10 @@ public byte readByte()
             return bytesStore.readByte(offset);
 
         } catch (BufferUnderflowException e) {
-            return 0;
+            if (lenient) {
+                return 0;
+            }
```
minborg, line 411: This is a massive performance killer...

minborg (approved):

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
JerryShea: You mean throwing and catching the exception? Agreed. But we should not be using `lenient == true` in the hot path

JerryShea (approved):

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
alamar: ...or run into EOS in the hot path

PR #334, opened 2022-02-22
Fix for JVM crash when writing trivially copyable to elastic buffer

glukos:
I've bumped into this while trying to support efficient `copyTo` for `TriviallyCopyableEvent`.
```
    default S unsafeWriteObject(Object o, int offset, int length)
        if (this.isDirectMemory()) {
            final long dest = addressForWrite(writePosition());
            writeSkip(length); // blow up here if this isn't going to work
            UnsafeMemory.MEMORY.copyMemory(o, offset, dest, length);
```
There's a bug: `writeSkip` may cause grow of underlying bytes store (in fact, grow creates a new store that replaces the old one), and `dest` becomes invalid.

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/StreamingDataOutput.java
+++ b/src/main/java/net/openhft/chronicle/bytes/StreamingDataOutput.java
@@ -533,8 +533,8 @@ public interface StreamingDataOutput<S extends StreamingDataOutput<S>> extends S
     default S unsafeWriteObject(Object o, int offset, int length)
             throws BufferOverflowException, IllegalStateException {
         if (this.isDirectMemory()) {
-            final long dest = addressForWrite(writePosition());
             writeSkip(length); // blow up here if this isn't going to work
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/bytes/StreamingDataOutput.java
```diff
@@ -533,8 +533,8 @@ default S unsafeWriteObject(Object o, int length)
     default S unsafeWriteObject(Object o, int offset, int length)
             throws BufferOverflowException, IllegalStateException {
         if (this.isDirectMemory()) {
-            final long dest = addressForWrite(writePosition());
             writeSkip(length); // blow up here if this isn't going to work
+            final long dest = addressForWrite(writePosition() - length);
```
JerryShea, line 537: nice one. If length > chunkSize then this can blow up, although I appreciate that is unlikely. Probably not worth the cost/benefit of checking

JerryShea (approved):

PR #343, opened 2022-03-21
Don't readInt with existing offset when setting BytesStore on AbstractReference

### LINKED ISSUE #342: AbstractReference uses existing offset to trigger initialisation of newly assigned BytesStore
When `AbstractReference#bytesStore(bytes, offset, length)` is called the reference calls `BytesStore#readInt` with its existing offset value. This is not guaranteed to be in the range for the new bytes.

This is the root cause of https://github.com/OpenHFT/Chronicle-Queue/issues/1025

nicktindall:
Fixes #342

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/ref/AbstractReference.java
+++ b/src/main/java/net/openhft/chronicle/bytes/ref/AbstractReference.java
@@ -70,7 +70,6 @@ public abstract class AbstractReference extends AbstractCloseable implements Byt
         if (this.bytes != null) {
             this.bytes.release(this);
         }
-        bytes.readInt(offset); // wake it up if it's not pointing to anything yet.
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/bytes/ref/AbstractReference.java
```diff
@@ -70,7 +70,6 @@ protected void acceptNewBytesStore(@NotNull final BytesStore bytes)
         if (this.bytes != null) {
             this.bytes.release(this);
         }
-        bytes.readInt(offset); // wake it up if it's not pointing to anything yet.
```
nicktindall, line 73: This actually seems to be called on line 50 anyway (with the correct offset), I think it's redundant here?

JerryShea (approved):

PR #344, opened 2022-03-25
Issue/318

minborg:
This PR addresses many of the outstanding SQ issues in Bytes. I really would like to get rid of them so we could see what to improve in the future.

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/Bytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/Bytes.java
@@ -1087,14 +1087,14 @@ public interface Bytes<U> extends
      * @return index of equal contents or -1
      * @throws NullPointerException if the provided {@code source} is {@code null}
      */
-    default long indexOf(@NotNull Bytes source)
+    default long indexOf(@NotNull final Bytes source)
             throws IllegalStateException {
         throwExceptionIfReleased(this);
         throwExceptionIfReleased(source);
-        long sourceOffset = readPosition();
-        long otherOffset = source.readPosition();
-        long sourceCount = readRemaining();
-        long otherCount = source.readRemaining();
+        final long sourceOffset = readPosition();
+        final long otherOffset = source.readPosition();
+        final long sourceCount = readRemaining();
+        final long otherCount = source.readRemaining();
 
         if (sourceCount <= 0) {
             return Math.toIntExact(otherCount == 0 ? sourceCount : -1);
@@ -1103,8 +1103,8 @@ public interface Bytes<U> extends
             return 0;
         }
         try {
-            byte firstByte = source.readByte(otherOffset);
-            long max = sourceOffset + (sourceCount - otherCount);
+            final byte firstByte = source.readByte(otherOffset);
+            final long max = sourceOffset + (sourceCount - otherCount);
 
             for (long i = sourceOffset; i <= max; i++) {
                 /* Look for first character. */
@@ -1115,7 +1115,7 @@ public interface Bytes<U> extends
                 /* Found first character, now look at the rest of v2 */
                 if (i <= max) {
                     long j = i + 1;
-                    long end = j + otherCount - 1;
+                    final long end = j + otherCount - 1;
                     for (long k = otherOffset + 1; j < end && readByte(j) == source.readByte(k); j++, k++) {
                         // Do nothing
                     }
@@ -1150,14 +1150,14 @@ public interface Bytes<U> extends
      * @return index of equal contents or -1
      * @throws NullPointerException if the provided {@code source } is {@code null}
      */
-    default int indexOf(@NotNull BytesStore source, int fromIndex)
-            throws IllegalStateException {
+    default int indexOf(@NotNull final BytesStore source,
+                        int fromIndex) throws IllegalStateException {
         throwExceptionIfReleased(this);
         throwExceptionIfReleased(source);
-        long sourceOffset = readPosition();
-        long otherOffset = source.readPosition();
-        long sourceCount = readRemaining();
-        long otherCount = source.readRemaining();
+        final long sourceOffset = readPosition();
+        final long otherOffset = source.readPosition();
+        final long sourceCount = readRemaining();
+        final long otherCount = source.readRemaining();
 
         if (fromIndex < 0) {
             fromIndex = 0;
@@ -1169,8 +1169,8 @@ public interface Bytes<U> extends
             return fromIndex;
         }
         try {
-            byte firstByte = source.readByte(otherOffset);
-            long max = sourceOffset + (sourceCount - otherCount);
+            final byte firstByte = source.readByte(otherOffset);
+            final long max = sourceOffset + (sourceCount - otherCount);
 
             for (long i = sourceOffset + fromIndex; i <= max; i++) {
                 /* Look for first character. */
@@ -1181,7 +1181,7 @@ public interface Bytes<U> extends
                 /* Found first character, now look at the rest of v2 */
                 if (i <= max) {
                     long j = i + 1;
-                    long end = j + otherCount - 1;
+                    final long end = j + otherCount - 1;
                     for (long k = otherOffset + 1; j < end && readByte(j) == source.readByte(k); j++, k++) {
                         // Do nothing
                     }
--- a/src/main/java/net/openhft/chronicle/bytes/Bytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/Bytes.java
@@ -1087,14 +1087,14 @@ public interface Bytes<U> extends
      * @return index of equal contents or -1
      * @throws NullPointerException if the provided {@code source} is {@code null}
      */
-    default long indexOf(@NotNull final Bytes source)
+    default long indexOf(@NotNull Bytes source)
             throws IllegalStateException {
         throwExceptionIfReleased(this);
         throwExceptionIfReleased(source);
-        final long sourceOffset = readPosition();
-        final long otherOffset = source.readPosition();
-        final long sourceCount = readRemaining();
-        final long otherCount = source.readRemaining();
+        long sourceOffset = readPosition();
+        long otherOffset = source.readPosition();
+        long sourceCount = readRemaining();
+        long otherCount = source.readRemaining();
 
         if (sourceCount <= 0) {
             return Math.toIntExact(otherCount == 0 ? sourceCount : -1);
@@ -1103,8 +1103,8 @@ public interface Bytes<U> extends
             return 0;
         }
         try {
-            final byte firstByte = source.readByte(otherOffset);
-            final long max = sourceOffset + (sourceCount - otherCount);
+            byte firstByte = source.readByte(otherOffset);
+            long max = sourceOffset + (sourceCount - otherCount);
 
             for (long i = sourceOffset; i <= max; i++) {
                 /* Look for first character. */
@@ -1115,7 +1115,7 @@ public interface Bytes<U> extends
                 /* Found first character, now look at the rest of v2 */
                 if (i <= max) {
                     long j = i + 1;
-                    final long end = j + otherCount - 1;
+                    long end = j + otherCount - 1;
                     for (long k = otherOffset + 1; j < end && readByte(j) == source.readByte(k); j++, k++) {
                         // Do nothing
                     }
@@ -1150,14 +1150,14 @@ public interface Bytes<U> extends
      * @return index of equal contents or -1
      * @throws NullPointerException if the provided {@code source } is {@code null}
      */
-    default int indexOf(@NotNull final BytesStore source,
-                        int fromIndex) throws IllegalStateException {
+    default int indexOf(@NotNull BytesStore source, int fromIndex)
+            throws IllegalStateException {
         throwExceptionIfReleased(this);
         throwExceptionIfReleased(source);
-        final long sourceOffset = readPosition();
-        final long otherOffset = source.readPosition();
-        final long sourceCount = readRemaining();
-        final long otherCount = source.readRemaining();
+        long sourceOffset = readPosition();
+        long otherOffset = source.readPosition();
+        long sourceCount = readRemaining();
+        long otherCount = source.readRemaining();
 
         if (fromIndex < 0) {
             fromIndex = 0;
@@ -1169,8 +1169,8 @@ public interface Bytes<U> extends
             return fromIndex;
         }
         try {
-            final byte firstByte = source.readByte(otherOffset);
-            final long max = sourceOffset + (sourceCount - otherCount);
+            byte firstByte = source.readByte(otherOffset);
+            long max = sourceOffset + (sourceCount - otherCount);
 
             for (long i = sourceOffset + fromIndex; i <= max; i++) {
                 /* Look for first character. */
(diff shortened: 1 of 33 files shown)


minborg:
I anticipate comments related to the size of this PR. I know it is big.

src/main/java/net/openhft/chronicle/bytes/Bytes.java
```diff
@@ -1087,14 +1087,14 @@ default BigInteger readBigInteger()
      * @return index of equal contents or -1
      * @throws NullPointerException if the provided {@code source} is {@code null}
      */
-    default long indexOf(@NotNull Bytes source)
+    default long indexOf(@NotNull final Bytes source)
```
alamar: I suggest we skip changes which only consist of setting final/NotNull/Nullable without adding any improvements to the code. I think we can use final as a matter of personal preference but the syntax is quite noisy (compared to var/val which we will probably have one day) so it does not make sense to apply this universally.

src/main/java/net/openhft/chronicle/bytes/Bytes.java
```diff
@@ -1150,14 +1150,14 @@ default long indexOf(@NotNull Bytes source)
      * @return index of equal contents or -1
      * @throws NullPointerException if the provided {@code source } is {@code null}
      */
-    default int indexOf(@NotNull BytesStore source, int fromIndex)
-            throws IllegalStateException {
+    default int indexOf(@NotNull final BytesStore source,
+                        int fromIndex) throws IllegalStateException {
```
alamar: In this case, it would make more sense to make fromIndex immutable rather than setting final on variables which are already immutable.

src/main/java/net/openhft/chronicle/bytes/BytesMarshaller.java
```diff
@@ -101,32 +102,37 @@ public static Object create(@NotNull Field field) {
...
-                    if (BytesStore.class.isAssignableFrom(type))
-                        return new BytesFieldAccess(field);
-                    if (BytesMarshallable.class.isAssignableFrom(type))
-                        return new BytesMarshallableFieldAccess(field);
-                    return new ScalarFieldAccess(field);
+                    return nonPrimitiveFieldAccess(field, type);
+            }
+        }
+
+        static Object nonPrimitiveFieldAccess(@NotNull final Field field,
```
alamar, line 108: Good catch.

src/main/java/net/openhft/chronicle/bytes/BytesMarshaller.java
```diff
@@ -345,7 +352,7 @@ else if (c.length != length)
         }
     }
 
-    static class CollectionFieldAccess extends FieldAccess {
+    static final class CollectionFieldAccess extends FieldAccess {
```
alamar: It makes sense to use final in anything exposed as APIs.

src/main/java/net/openhft/chronicle/bytes/BytesUtil.java
```diff
@@ -54,12 +54,12 @@
      * @param clazz to check
      * @return true if the whole class is trivially copyable
      */
-    public static boolean isTriviallyCopyable(Class clazz) {
-        int[] ints = TRIVIALLY_COPYABLE.get(clazz);
+    public static boolean isTriviallyCopyable(@NotNull final Class<?> clazz) {
```
alamar: The method is public - it would be nice if we moved the whole class to `internal`/`impl` package to discourage any potential use by library end user.

alamar: Maybe also de-enumize it

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
```diff
@@ -69,7 +69,8 @@ public HexDumpBytes() {
         }
     }
 
-    HexDumpBytes(@NotNull NativeBytes<Void> base, @NotNull BytesStore text) {
+    HexDumpBytes(@NotNull final NativeBytes<Void> base,
```
alamar: I don't see why reformat it, as long if it fits into 100 characters limit.

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
```diff
@@ -790,109 +793,112 @@ public String read8bit()
...
-    public <C extends Appendable & CharSequence> long readUtf8(long offset, @NotNull C sb)
+    public <C extends Appendable & CharSequence> long readUtf8(final long offset,
+                                                               @NotNull final C sb)
             throws IORuntimeException, IllegalArgumentException, BufferUnderflowException, ArithmeticException, IllegalStateException {
         return base.readUtf8(offset, sb);
     }
 
     @Override
-    public <C extends Appendable & CharSequence> long readUtf8Limited(long offset, @NotNull C sb, int maxUtf8Len)
+    public <C extends Appendable & CharSequence> long readUtf8Limited(final long offset,
```
alamar: For better readability, maybe a newline before "long" and smallrt indent.

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
```diff
@@ -945,46 +951,54 @@ public long writePosition() {
...
-                    if (wp > 0 && text.peekUnsignedByte(wp - 1) > ' ')
-                        text.append(' ');
-                    try {
-                        text.appendBase16(value, 2);
-                    } catch (IllegalArgumentException | BufferOverflowException e) {
-                        throw new AssertionError(e);
-                    }
-                } while (pos < end);
-            }
+            performCopyToText(pos);
```
alamar: copyToText0 perhaps?

src/main/java/net/openhft/chronicle/bytes/MappedBytes.java
```diff
@@ -66,8 +66,9 @@ public static MappedBytes singleMappedBytes(@NotNull final File file, final long
     }
 
     @NotNull
-    public static MappedBytes singleMappedBytes(@NotNull final File file, final long capacity, boolean readOnly)
-            throws FileNotFoundException, IllegalStateException {
+    public static MappedBytes singleMappedBytes(@NotNull final File file,
+                                                final long capacity,
+                                                final boolean readOnly) throws FileNotFoundException, IllegalStateException {
```
alamar: Line becomes too long.

JerryShea (dismissed):
final classes!

src/main/java/net/openhft/chronicle/bytes/NativeBytes.java
```diff
@@ -300,6 +302,17 @@ private void resize(final long endOfBuffer)
         }
     }
 
+    @NotNull
+    private BytesStore allocate(long size) {
+        final BytesStore store;
+        try {
+            store = allocateNewByteBufferBackedStore(Maths.toInt32(size));
```
alamar, line 309: Maybe bake the check in allocateNewByteBufferBackedStore

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -1739,47 +1760,6 @@ private static void appendLong0(@NotNull StreamingDataOutput out, long num)
         }
     }
 
-    private static int appendInt1(byte[] numberBuffer, int num) {
```
alamar, line 1742: Dead code? Nice.

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedBytes.java
```diff
@@ -153,7 +153,7 @@ private long copySize(final long writePosition) {
     @Override
     public Bytes<Void> readPositionRemaining(final long position, final long remaining)
             throws BufferUnderflowException, IllegalStateException {
-//        throwExceptionIfClosed();
+        // throwExceptionIfClosed
```
alamar: I think these may be removed altogether.

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedFile.java
```diff
@@ -180,7 +187,7 @@ public MappedBytesStore acquireByteStore(
             }
             // *** THIS CAN TAKE A LONG TIME IF A RESIZE HAS TO OCCUR ***
             // let double check it to make sure no other thread change it in the meantime.
-            //resizeRafIfTooSmall(chunk);
+            // resizeRafIfTooSmall ( chunk )
```
alamar: I wonder what that means.

src/main/java/net/openhft/chronicle/bytes/internal/migration/HashCodeEqualsUtil.java
```diff
@@ -5,17 +5,15 @@
...
 
     private HashCodeEqualsUtil() {
     }
 
     // Comparing and reserving at the same time
-    // Declared private at the moment so not public
-    private static boolean contentEquals(final BytesStore<?, ?> bytes,
-                                        final Object other) {
+    // Declared package private at the moment so not public
+    static boolean contentEquals(final BytesStore<?, ?> bytes,
```
alamar: Why? Are there plans to use it?

src/test/java/net/openhft/chronicle/bytes/MappedBytesTest.java
```diff
@@ -256,107 +256,43 @@ public void testWriteReadBytesWithOffsetAndTextShift()
...
+            assertTrue(ex.getMessage().startsWith("Acquired the next BytesStore"));
+        }
+    }
 
-        bytes2.releaseLast();
-        bytes.releaseLast();
 
-        assertTrue(bytes.isClosed());
+    @Test
+    public void testLargeWrites() throws IOException {
```
alamar, line 274: That's a lot of copy-paste!

src/main/java/net/openhft/chronicle/bytes/BytesUtil.java
minborg: We could do this as a separate effort. There are many other classes like this

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
minborg: I find it easier to have one parameter per row but this is a matter of taste.

minborg:
> final classes!

Yes. Great progress. I was able to mark these classes as `final` as they were package private and were not overridden.

src/main/java/net/openhft/chronicle/bytes/Bytes.java
minborg: Let's discuss this at the next dev meeting. I have only updated the code that I touched.

JerryShea: Agree @alamar

minborg: Do you think we should roll back the changes with respect to this?

src/main/java/net/openhft/chronicle/bytes/Bytes.java
minborg: How do you mean? ByteStore is mutable. `source` is not updated within the method.

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedFile.java
minborg: Me too

src/main/java/net/openhft/chronicle/bytes/internal/migration/HashCodeEqualsUtil.java
minborg: This provides a conentsEqual that makes sure the ByteStores are not released during execution. However, benchmarks show is was slow. Do you think we should remove it?

src/test/java/net/openhft/chronicle/bytes/MappedBytesTest.java
minborg: Yepp

src/main/java/net/openhft/chronicle/bytes/DistributedUniqueTimeProvider.java
```diff
@@ -142,7 +142,10 @@ public static long hostIdFor(long timestampWithHostId) {
         return timestampWithHostId % HOST_IDS;
     }
 
-    static class DistributedUniqueTimeProviderHolder {
+    static final class DistributedUniqueTimeProviderHolder {
```
minborg, line 145: This is a "utility class" and is thus final and has a private constructor

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
minborg: I've tried different versions and no one looks really good to me. The most annoying thing is that the method declares RuntimeException variants.

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
minborg: Sounds good

src/main/java/net/openhft/chronicle/bytes/MappedBytes.java
minborg: Yes. Again the problem with IllegalStateException. I have fixed it

src/main/java/net/openhft/chronicle/bytes/NativeBytes.java
minborg: The purpose here was to get rid of nested try/catch

src/main/java/net/openhft/chronicle/bytes/Bytes.java
alamar: I would prefer that method changes which only have final added, or notnull added when they are not directly exposed as API, to be rolled back. This one seems to be the prime example.

src/main/java/net/openhft/chronicle/bytes/Bytes.java
alamar: fromIndex is a parameter and it is a code smell to modify a parameter's value. More so than not having final/NotNull.

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
alamar: I don't think you should reformat code unless you touch it in any other way.

src/main/java/net/openhft/chronicle/bytes/NativeBytes.java
alamar: Is there a chance we can move this try/catch into allocateNewByteBufferBackedStore()?

src/main/java/net/openhft/chronicle/bytes/internal/migration/HashCodeEqualsUtil.java
alamar: Why do we need to expose it in package private fashion if it wasn't previously?

minborg:
> final classes!

I have rolled back the final classes

src/main/java/net/openhft/chronicle/bytes/Bytes.java
minborg: I totally agree. But there is already too much in this PR so I opted not the change it now.

src/main/java/net/openhft/chronicle/bytes/internal/migration/HashCodeEqualsUtil.java
minborg: I will remove it

alamar (approved):

src/main/java/net/openhft/chronicle/bytes/internal/migration/HashCodeEqualsUtil.java
peter-lawrey: It will be slow when tracing is enabled as it is by default in this module. It shouldn't be slow when this is off.

PR #347, opened 2022-03-28
Fixsqwarn

minborg:


src/main/java/net/openhft/chronicle/bytes/MappedFile.java
```diff
@@ -283,6 +283,7 @@ public void setNewChunkListener(final NewChunkListener listener) {
     @NotNull
     public abstract RandomAccessFile raf();
 
+    @Deprecated
```
alamar, line 291: Should we add comment "To be removed in 2.2n"?

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedFile.java
```diff
@@ -358,6 +358,7 @@ public RandomAccessFile raf() {
         return raf;
     }
 
+    @Deprecated
```
alamar, line 366: ditto

alamar (changes requested):

src/main/java/net/openhft/chronicle/bytes/MappedFile.java
minborg: The removal of the override is a bit complicated. We need to move from `finalize()` to perhaps using `WeakReference` or something else. The end-user should never invoke this method so I do not think we have to say when.

minborg: In other words, it is not a part of the public API.

alamar (approved):

PR #352, opened 2022-03-28
Specify warmup dir for ChunkedMappedFile::warmup, Fix #351

### LINKED ISSUE #351: Specify warmup dir for ChunkedMappedFile::warmup
It is a security issue to place a file in tmp.

minborg:


src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedFile.java
```diff
@@ -84,7 +85,7 @@ public static void warmup() {
         try {
             Jvm.setExceptionHandlers(Slf4jExceptionHandler.ERROR, null, null);
 
-            final Path path = Files.createTempDirectory("warmup");
+            final Path path = Files.createTempDirectory(Paths.get("warmup_dir"), "warmup");
```
alamar, line 88: Should it be configurable by property? Will it be allocated in tmpdir or curdir?

minborg: This creates a directory in the systems's default tmp directory (e.g. `/tmp` or `\Windows\Temp`). We could very easily add a system property. But it needs to be documented too.

minborg:
We are not acting on any read data so it is safe to mark this vulnerability as "safe".

PR #353, opened 2022-03-28
Hoist out boundary checks, Fix #297

### LINKED ISSUE #297: Add other methods with hoisted boundary checks
For example `startsWith` and 'compareUtf8' in `BytesInternal`

minborg:


src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -319,37 +319,75 @@ public static boolean startsWith(@NotNull BytesStore a, @NotNull BytesStore b)
...
-                i += 4;
-            }
-            if (i < length - 1) {
-                if (a.readShort(aPos + i) != b.readShort(bPos + i))
-                    return false;
-                i += 2;
-            }
-            if (i < length) {
-                return a.readByte(aPos + i) == b.readByte(bPos + i);
+            if (a instanceof HasUncheckedRandomDataInput && b instanceof HasUncheckedRandomDataInput) {
```
minborg: We could add a check for a certain length threshold here before we do the intstanceof.

nicktindall: Would it be faster if the interface had a startsWith method, then you'd only have to `instanceof` the RHS? (The LHS would just use whatever implementation it implemented)

E.g. a.startsWith(b)

Ignore me if this is gibberish.

nicktindall (approved):

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
minborg: The best would be if we had "traits" and then we could just select two different views: one checked and one unchecked. I.e. we separate the concerns of memory and access.

The solution above only caters to the case both can be optimized when, in reality, there are four combinations. The other three will use checked access.

Related: https://stackoverflow.com/questions/103564/the-performance-impact-of-using-instanceof-in-java

minborg: But your suggestion would work and we could override the relevant classes. It would indeed mean one less `instanceof `. There is already an implementation in each concrete class (inherited from `BytesStore`)

alamar: You can just add a method overload, i.e. public static boolean startsWith(@NotNull HasUncheckedRandomDataInput a, @NotNull HasUncheckedRandomDataInput b) - and then optionally call it here.

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -319,37 +319,75 @@ public static boolean startsWith(@NotNull BytesStore a, @NotNull BytesStore b)
...
     }
 
+    public static boolean startsWithUnchecked(@NotNull final BytesStore a,
+                                              @NotNull final BytesStore b,
+                                              @NonNegative final long aPos,
+                                              @NonNegative final long bPos,
+                                              @NonNegative final long length) {
+        final UncheckedRandomDataInput ua = ((HasUncheckedRandomDataInput) a).acquireUncheckedInput();
+        final UncheckedRandomDataInput ub = ((HasUncheckedRandomDataInput) b).acquireUncheckedInput();
+        int i;
```
alamar, line 363: Some explanation is expected (in comments perhaps, or in the form of @see)

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -319,37 +319,75 @@ public static boolean startsWith(@NotNull BytesStore a, @NotNull BytesStore b)
...
             throw new AssertionError(e);
         }
     }
 
+    public static boolean startsWithUnchecked(@NotNull final BytesStore a,
+                                              @NotNull final BytesStore b,
+                                              @NonNegative final long aPos,
+                                              @NonNegative final long bPos,
+                                              @NonNegative final long length) {
+        final UncheckedRandomDataInput ua = ((HasUncheckedRandomDataInput) a).acquireUncheckedInput();
```
alamar: You can specify correct type for a&b parameters and do casts elsewhere.

PR #354, opened 2022-03-31
Zeroassert

minborg:


.gitignore
```diff
@@ -6,7 +6,6 @@
 *.class
 
 ### Package Files
-*.jar
```
alamar: This change should now be reverted.

pom.xml
```diff
@@ -311,6 +319,17 @@
     </scm>
 
     <profiles>
+
+        <profile>
+            <id>enable-zero-cost-assertions</id>
+            <dependencies>
+                <dependency>
+                    <groupId>net.openhft</groupId>
+                    <artifactId>chronicle-core-assertions</artifactId>
```
alamar: This change should now be reverted

src/main/java/net/openhft/chronicle/bytes/internal/ZeroCostSupport.java
```diff
@@ -0,0 +1,11 @@
+package net.openhft.chronicle.bytes.internal;
+
+import net.openhft.chronicle.core.util.AssertUtil;
+
+class ZeroCostSupport {
+
+    boolean isZeroCostAssertionsEnabled() {
+        return !AssertUtil.SKIP_ASSERTIONS;
+    }
```
alamar: Do we need it?

src/test/java/net/openhft/chronicle/bytes/internal/ZeroCostSupportTest.java
```diff
@@ -0,0 +1,11 @@
+package net.openhft.chronicle.bytes.internal;
+
+import org.junit.jupiter.api.Test;
+
+class ZeroCostSupportTest {
+
+    @Test
+    void isZeroCostAssertionsEnabled() {
```
alamar: Ditto

alamar (changes requested):

alamar (approved):

Issue #358, opened 2022-04-12
Make Bytes, BytesIn and ByteOut generic to fix raw type warning


peter-lawrey:
Some projects now fail when warning about raw types are produced by the compiler.

PR #359, opened 2022-04-13
Make Bytes, BytesIn and ByteOut generic to fix raw type warning https://github.com/OpenHFT/Chronicle-Bytes/issues/358

JerryShea:


src/main/java/net/openhft/chronicle/bytes/BytesMethodReaderBuilder.java
```diff
@@ -24,19 +24,19 @@
 
 @SuppressWarnings("rawtypes")
```
minborg: Remove suppression?

src/main/java/net/openhft/chronicle/bytes/DistributedUniqueTimeProvider.java
```diff
@@ -61,7 +61,7 @@ public static DistributedUniqueTimeProvider instance() {
     }
 
     @SuppressWarnings("rawtypes")
```
minborg: ditto

src/main/java/net/openhft/chronicle/bytes/MappedBytes.java
```diff
@@ -31,9 +31,9 @@
 import java.io.FileNotFoundException;
 
 /**
- * Bytes to wrap memory mapped data.
+ * Bytes<?> to wrap memory mapped data.
  * <p>
- * NOTE These Bytes are single Threaded as are all Bytes.
+ * NOTE These Bytes<?> are single Threaded as are all Bytes.
  */
 @SuppressWarnings({"rawtypes", "unchecked"})
```
minborg: remove "rawtypes"?

minborg:
I am very pleased to see this PR. Could we go through the code and try to remove as many suppressed warnings of type "rawtypes" as possible?

Some of the changes may lead to compatibility issues but in many of those cases, there is an underlying problem anyhow that should be flushed out.

minborg (approved):

src/main/java/net/openhft/chronicle/bytes/NativeBytes.java
```diff
@@ -40,7 +40,7 @@
 /**
  * Elastic memory accessor which can wrap either a ByteBuffer or malloc'ed memory.
  * <p>
- * <p>This class can wrap <i>heap</i> ByteBuffers, called <i>Native</i>Bytes for historical reasons.
+ * <p>This class can wrap <i>heap</i> ByteBuffers, called <i>Native</i>Bytes<?> for historical reasons.
```
JerryShea: this is picky but it would be good if we could undo the changes in comments

PR #360, opened 2022-05-04
deprecating mutating methods from readonly interfaces #967

vach-chronicle:


src/main/java/net/openhft/chronicle/bytes/RandomCommon.java
```diff
@@ -216,9 +216,11 @@ Bytes<?> bytesForWrite()
      * @param value    to set
      * @return true, if successful.
      */
+    @Deprecated(/* to be removed in x.25 */)
```
JerryShea: Can we say something like `@Deprecated(/* Use this_other_interface_instead, to be removed in x.25 */)` i.e. please tell the user what to use instead

JerryShea: Are these methods implement somewhere else e.g. on one of the output interfaces?

vach-chronicle: yes, they are going to be in output interfaces, however if I add them now, they still show up as deprecated
let me add it since there is no harm
and I'll add more descriptive comment

minborg:
Please review the SonarCloud code smells above.

src/main/java/net/openhft/chronicle/bytes/RandomDataOutput.java
```diff
@@ -31,7 +31,7 @@
 import static net.openhft.chronicle.core.util.ObjectUtils.requireNonNull;
 
 @SuppressWarnings({"rawtypes", "unchecked"})
-public interface RandomDataOutput<R extends RandomDataOutput<R>> extends RandomCommon {
+public interface RandomDataOutput<R extends RandomDataOutput<R>> extends RandomDataInput, RandomCommon {
```
JerryShea, line 34: sure you want to extend RandomDataInput?

vach-chronicle: Yes, this is somewhat conceptual change if you have access to write you should have access to read as well, for example in order to do CAS you need both... but that's not the reason I do it this way.

I would compare this to List vs Immutable versions of it, in java this sucks because both are shared around with LIst interface, but for example, in kotlin we have List (which has read-only methods) and MutableList which has read+write (equivalent to java List)... 

It would not make much sense to create a List that has only write methods, you'd have to blindly write to it, what if its a linked list and you want to peek... 

It's the same concept here. If you have to write access to buffer you should have read as well, if you have read access you should not be able to write.

minborg: There could be an underlying source (like Wire) and then we could derive two different views of the underlying source: one for reading and one for writing. The latter can only be used for writing but will mutate the underlying source. These mutations can be picked up bu the reading view and directly from the underlying source.

vach-chronicle: Had a discussion with Per, 
and will post a separate PR that addresses this change with the suggestions that he made
I think it will boil down to what is conceptually preferable?
this PR = Reader can only read, the writer can both read and write
the other PR = Reader only reads, Writers can only write, if you want both (do a typecast)

minborg: > Had a discussion with Per, and will post a separate PR that addresses this change with the suggestions that he made I think it will boil down to what is conceptually preferable? this PR = Reader can only read, the writer can both read and write the other PR = Reader only reads, Writers can only write, if you want both (do a typecast)

do a typecast or use an entity that implements both

vach-chronicle: Ok just to summerise, there are 2 suggestions for this change

1. (this PR) If you have a write access to a buffer you should also be able to read. This is a temporary branch with this approach applied https://github.com/OpenHFT/Chronicle-Bytes/commit/362132af14d15dfa7eb4ef2917bc9cf955bc1758.

2. (as Per suggested https://github.com/OpenHFT/Chronicle-Bytes/pull/362) Write interface for buffer stays write-only (as it was originally created) but when you need to read as well you have to do typecast... this means user code as well must do a typecast like here https://github.com/OpenHFT/Chronicle-Bytes/commit/8d224d85c7f0e660f1961debe56741f605d3b993)

JerryShea: I think my vote is for https://github.com/OpenHFT/Chronicle-Bytes/commit/8d224d85c7f0e660f1961debe56741f605d3b993

vach-chronicle: ok then please @JerryShea @minborg look at this PR - https://github.com/OpenHFT/Chronicle-Bytes/pull/362
i'll close this one later

vach-chronicle:
alternative solution was merged, closing @minborg

PR #362, opened 2022-05-11
deprecating mutating methods from readonly interfaces #967

vach-chronicle:
This is an alternative version of interface change, we keep writers write only and readers read only
however we will have to do type casts whenever we need to read while having Writer interface (this is how deprecations would get applied https://github.com/OpenHFT/Chronicle-Bytes/commit/8d224d85c7f0e660f1961debe56741f605d3b993)

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/RandomCommon.java
+++ b/src/main/java/net/openhft/chronicle/bytes/RandomCommon.java
@@ -215,9 +215,11 @@ interface RandomCommon extends ReferenceCounted {
      * @param value    to set
      * @return true, if successful.
      */
+    @Deprecated(/* Use RandomDataOutput instead, to be removed in x.25 */)
     boolean compareAndSwapInt(@NonNegative long offset, int expected, int value)
             throws BufferOverflowException, IllegalStateException;
 
+    @Deprecated(/* Use RandomDataOutput instead, to be removed in x.25 */)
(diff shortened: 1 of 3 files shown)


minborg (approved):

src/main/java/net/openhft/chronicle/bytes/RandomCommon.java
```diff
@@ -240,6 +243,7 @@ boolean compareAndSwapLong(@NonNegative long offset, long expected, long value)
      * @param value    to set
      * @return true, if successful.
      */
+    @Deprecated(/* Use RandomDataOutput instead, to be removed in x.26 */)
```
JerryShea: we are now x.23 so can remove in x.25

src/main/java/net/openhft/chronicle/bytes/RandomDataOutput.java
```diff
@@ -433,4 +433,5 @@ default long writeUtf8Limited(@NonNegative long writeOffset, @Nullable CharSeque
 
     long write8bit(@NonNegative long position, @NotNull String s, @NonNegative int start, @NonNegative int length);
 
+    // TODO move deprecated mutating methods from RandomCommon and RandomDataInput x.26
```
JerryShea: don't we have to implement these TODOs before we can merge?

PR #378, opened 2022-05-31
Replace NoByteStore, Fix #374

### LINKED ISSUE #374: Improve NoBytesStore
There are some problems with this class:

- [ ] `NoBytesStore` should be immutable but shares a mutable state via `addressForRead()`, `addressForWrite()` and `addressForWritePosition()`
- [ ] It is using raw types
- [ ] `copy()` does not return a copy and copy is not fully specified
- [ ] `bytesForRead()` returns a new `Bytes` object whereas `bytesForWrite()` does not
- [ ] it is not tested
- [ ] `copyTo(@NotNull OutputStream out)` is not overridden
- [ ] `zeroOut` is not overridden
- [ ] The class javadoc says it can be resized
- [ ] and more...

minborg:
This fixes a number of problems with the old NoByteStore and reduces the API.


src/main/java/net/openhft/chronicle/bytes/internal/SingletonEmptyByteStore.java
```diff
@@ -0,0 +1,505 @@
...
+
+    @NotNull
+    @Override
+    public EmptyByteStore copy() {
+        return this;
+    }
+
+    @Override
+    public @NotNull Bytes<Void> bytesForRead() throws IllegalStateException {
+        return new VanillaBytes<Void>(this, 0, 0) {
```
alamar: I think this constructor should only be used by descendants of VanillaBytes, can we not push the envelope here?

minborg: Conveniently, there is a `VanillaBytes.vanillaBytes()` that does this. I will fix it...

minborg: I suppose we could make the `VanillaBytes` class protected as it only has protected constructors. But that is for another time.

alamar (approved):

PR #385, opened 2022-06-02
Create a new ByteStore when writing zero sized arrays. Fix #384

### LINKED ISSUE #384: Bytes: Writing an array of length zero does not create new ByteStore
There is a shared singleton for empty ByteStores. However, there are cases where a real, unshared, ByteStore is assumed to exist. Therefore, even writing an array of length zero should trigger a replacement of the shared singleton.

minborg:


src/main/java/net/openhft/chronicle/bytes/NativeBytes.java
```diff
@@ -160,7 +161,8 @@ protected void writeCheckOffset(final @NonNegative long offset, final @NonNegati
             throws BufferOverflowException, IllegalStateException {
         if (offset >= bytesStore.start() && offset + adding >= bytesStore.start()) {
             final long writeEnd = offset + adding;
-            if (writeEnd <= bytesStore.safeLimit()) {
+            // Always resize if we are backed by a SingletonEmptyByteStore as this is shared and does not provide all functionality
+            if (writeEnd <= bytesStore.safeLimit() && !(EmptyByteStore.class.isInstance(bytesStore))) {
```
alamar: Suspicious approach, I think we should override this method in EmptyByteStore and remove this check altogether, without touching this method. Same below.

alamar (changes requested):
see inline

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -1321,6 +1322,17 @@ public int byteCheckSum(@NonNegative int start, @NonNegative int end)
         return sum & 0xFF;
     }
 
+    protected boolean isBackedByEmptyByteStore() {
+        return EmptyByteStore.class.isInstance(bytesStore);
```
alamar: I think it would be more performant (and certainly more object-oriented) if ByteStore had a default isEmptyBytesStore() returning false and EmptyByteStore had it overridden and return true - then you may remove this method

minborg: The problem is that AbstractBytes is in the public API available for anyone to override. But we could certainly override it for some implementations.

minborg: It is hard to override this method in any Bytes implementation as many are elastic and one can provide any initial ByteStore.

alamar: If it's a default method with sensible default, no action is required from implementors.

src/main/java/net/openhft/chronicle/bytes/NativeBytes.java
```diff
@@ -45,6 +47,9 @@
 @SuppressWarnings({"rawtypes", "unchecked"})
 public class NativeBytes<U>
         extends VanillaBytes<U> {
+
```
alamar: extra line feeds

alamar (approved):
I assume we can't make interface method protected? Approved.

minborg:
> I assume we can't make interface method protected? Approved.

I had the same thought.  Here is an article I wrote some time ago about it: https://dzone.com/articles/java-8-declare-private-and-protected-methods-in-in

In Java 9, you can.

PR #386, opened 2022-06-05
Issue 377

j4sm1ne96:
Add systemProperties.adoc

### DIFF
--- a/systemProperties.adoc
+++ b/systemProperties.adoc
@@ -1,0 +1,16 @@
+
+== System Properties
+Below, a number of relevant System Properties are listed.
+
+.System properties
+[cols=4*, options="header"]
+|===
+| Property Key | Default | Description | Java Variable Name (Type)
+| bytes.guarded | `false` | If 'true', @param guarded turn on | _BYTES_GUARDED_ (boolean)
+| bytes.bounds.unchecked | `false` | | _BYTES_BOUNDS_UNCHECKED_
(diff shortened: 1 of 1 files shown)


systemProperties.adoc
```diff
@@ -0,0 +1,22 @@
+
+== System Properties
+Below, a number of relevant System Properties are listed.
+
+NOTE:
+With boolean properties, see javadoc:
+`net.openhft.chronicle.core.Jvm#getBoolean(java.lang.String)`
+
+When boolean system properties are set to "", "yes" or "true", the property is enabled.
```
rogersimmons: "" - really?
An empty value for a boolean should be an error
@JerryShea - is this expected (or required) behaviour?

JerryShea: I agree that we should not be talking about "" - please see javadoc for net.openhft.chronicle.core.Jvm#getBoolean(java.lang.String, boolean)

We should link to this, and maybe consider pasting some of it in here

rogersimmons (approved):

systemProperties.adoc
peter-lawrey: Currently `-DskipTests` in maven is the same as `-DskipTests=` and `-DskipTests=true`
Are you suggesting the program should fail in the case of `-DskipTest=` ?

peter-lawrey: The code paths for `-DskipTest` and `-DskipTest=` are the same

Issue #294, opened 2021-11-18
Vectorize ByteInternal::contentEquals and other methods


minborg:
It might be possible to leverage `ArraysSupport::vectorizedMismatch` to improve performance.

Today, 8 bytes (64 bits) are accessed and compared as a unit rather than individually, which increases the performance when the method is compiled by the JVM.

Using _vectorized methods_, on supported platforms, the above implementation is intrinsified to leverage SIMD instructions. So for a memory segment, 16 bytes (128 bits), 32 bytes (256 bits), and perhaps in the future even 64 bytes (512 bits), platform permitting, can be accessed and compared as a unit, which further increases the performance over the Java implementation.

PR #387, opened 2022-06-06
Issue/294 Vectorize ByteInternal::contentEquals

RobAustin:
relatest to #294


pom.xml
```diff
@@ -391,6 +391,29 @@
...
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <argLine>--add-exports java.base/jdk.internal.ref=ALL-UNNAMED --add-exports
+                                java.base/jdk.internal.misc=ALL-UNNAMED --add-exports java.base/sun.nio.ch=ALL-UNNAMED
+                                --illegal-access=permit
+                                --add-exports java.base/jdk.internal.ref=ALL-UNNAMED
+                                --add-exports java.base/jdk.internal.util=ALL-UNNAMED
```
minborg: We should add this to the base pom so that libs and apps can take run automatically.

minborg: We should also add it to https://chronicle.software/chronicle-support-java-17/

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -95,13 +100,26 @@ enum BytesInternal {
     private static final int MAX_STRING_LEN = Jvm.getInteger("bytes.max-string-len", 128 * 1024);
     private static final int NEG_ONE = ByteOrder.nativeOrder() == ByteOrder.LITTLE_ENDIAN ? 0x80 : 0x8000;
 
+    static MethodHandle VECTORIZED_MISMATCH_METHOD_HANDLE;
```
minborg: If MHs are made `static final`, they can run at native speed.

minborg: So, we could add a static method that returns this MH .

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -118,11 +136,51 @@ public static boolean contentEqual(@Nullable final BytesStore a,
...
+        try {
+            long aAddress = left.addressForRead(left.readPosition());
+            long bAddress = right.addressForRead(right.readPosition());
+
+            return (int) VECTORIZED_MISMATCH_METHOD_HANDLE.invoke(
+                    null, aAddress,
+                    null, bAddress,
+                    (int) left.readRemaining(), 0) < 0;
+        } catch (Throwable e) {
+            throw new AssertionError(e);
```
minborg: I would have wrapped the Throwable into something that extends `RuntimeException`.

minborg:
This is really cool stuff! I wonder if it could be done for Objects too by putting in the appropriate distance to the pointer to continuous primitive array memory? Is not the array object and the actual array placed in different places?

pom.xml
JerryShea: This is already in base parent/root pom, and should be automatically enabled. What is different compared to https://github.com/OpenHFT/OpenHFT/blob/c38413e1f1c01bf893ad40e60e577ddccb674b5a/root-parent-pom/pom.xml#L399 ?

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -166,13 +170,21 @@ public static boolean contentEqual(@Nullable final BytesStore a,
...
+            System.out.println(((NativeBytes)right).toHexString());
+
+            System.out.println("left.realReadRemaining() = " + left.realReadRemaining());
+            System.out.println("right.realReadRemaining() = " + right.realReadRemaining());
+*/
             long aAddress = left.addressForRead(left.readPosition());
             long bAddress = right.addressForRead(right.readPosition());
 
-            return (int) VECTORIZED_MISMATCH_METHOD_HANDLE.invoke(
+            int invoke = (int) VECTORIZED_MISMATCH_METHOD_HANDLE.invoke(
```
minborg: Declaring primitive local variables `final` can improve performance. Check with `javap` and see the difference.

minborg: A general comment: the (int) cast Rob used above prevents auto-boxing.

pom.xml
RobAustin: > We should add this to the base pom so that libs and apps can take run automatically.

it was already added to the base pom and will be included on the base pom is released

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
RobAustin: ok

src/test/java/net/openhft/chronicle/bytes/internal/BytesInternalTest.java
```diff
@@ -0,0 +1,71 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(Parameterized.class)
+public class BytesInternalTest {
+    private final Bytes a;
+    private final Bytes b;
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> data() {
```
minborg, line 25: This could look more elegant under JUnit5. But leave it as is for now.

minborg (approved):

pom.xml
```diff
@@ -232,7 +238,7 @@
                     <source>1.8</source>
                     <target>1.8</target>
                     <encoding>UTF-8</encoding>
-<!--                    <verbose>true</verbose>-->
+                    <!--                    <verbose>true</verbose>-->
```
JerryShea, line 243: can we just delete this line?

pom.xml
```diff
@@ -331,11 +337,11 @@
                             <excludes>
                                 <!--<exclude>META-INF/**</exclude>-->
                                 <exclude>*.jar</exclude>
-<!--                                <exclude>net/openhft/chronicle/bytes/AbstractBytes.class</exclude>
-                                <exclude>net/openhft/chronicle/bytes/AppendableUtil.class</exclude>
-                                <exclude>net/openhft/chronicle/bytes/BinaryBytesMethodWriterInvocationHandler.class</exclude>
-                                <exclude>net/openhft/chronicle/bytes/BinaryWireCode.class</exclude>
-                                <exclude>net/openhft/chronicle/bytes/*.class</exclude>-->
+                                <!--                                <exclude>net/openhft/chronicle/bytes/AbstractBytes.class</exclude>
```
JerryShea, line 377: likewise, can we delete?

pom.xml
JerryShea: why not just release the root-parent-pom then we can just use that?

src/test/java/net/openhft/chronicle/bytes/ContentEqualsJLBHTest.java
```diff
@@ -0,0 +1,73 @@
+package net.openhft.chronicle.bytes;
+
+import net.openhft.chronicle.bytes.internal.BytesInternal;
+import net.openhft.chronicle.jlbh.JLBH;
+import net.openhft.chronicle.jlbh.JLBHOptions;
+import net.openhft.chronicle.jlbh.JLBHTask;
+
+public class ContentEqualsJLBHTest {
```
JerryShea, line 8: cool. We should add to `run-benchmarks` profile so we can see in TC graphs. You will also need to add
```
    @Override
    public void complete() {
        TeamCityHelper.teamCityStatsLastRun(this.getClass().getSimpleName(), jlbh, iterations, System.out);
    }
```

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -118,11 +149,90 @@ public static boolean contentEqual(@Nullable final BytesStore a,
...
+     * see https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8136924
+     * JDK-8033148 will add methods to Arrays for array equals, compare and mismatch.
+     * <p>
+     * The implementations of equals, compare and mismatch can be reimplemented using underlying mismatch methods that in turn defer to a single method, vectorizedMismatch, that accesses the memory contents of arrays using Unsafe.getLongUnaligned.
+     * <p>
+     * The vectorizedMismatch implementation can be optimized efficiently by C2 to obtain an approximate 8x speed up when performing a mismatch on byte[] arrays (of a suitable size to overcome fixed costs).
+     * <p>
+     * The contract of vectorizedMismatch is simple enough that it can be made an intrinsic (see JDK-8044082) and leverage SIMDs instructions to perform operations up to a width of say 512 bits on supported architectures. Thus even further performance improvements may be possible.
+     */
+
```
JerryShea, line 182: should not have blank lines between javadoc and method

src/test/java/net/openhft/chronicle/bytes/ContentEqualsJLBHTest.java
JerryShea: It would be good if this test could execute the old way and also the new way so we can see the difference

pom.xml
RobAustin: yes - we will

src/test/java/net/openhft/chronicle/bytes/ContentEqualsJLBHTest.java
RobAustin: I'll send you the email with the results.

PR #390, opened 2022-06-14
Fix test on Java 11 by removing reliance on try/catch #294

alamar:


alamar:
I got rid of try/catch approach since it may produce garbage on critical pathway.

RobAustin (approved):

JerryShea (approved):

PR #393, opened 2022-06-21
consistent thread safety check property names

JerryShea:
https://github.com/OpenHFT/Chronicle-Core/issues/409

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
@@ -28,10 +28,7 @@ import net.openhft.chronicle.core.Jvm;
 import net.openhft.chronicle.core.UnsafeMemory;
 import net.openhft.chronicle.core.annotation.NonNegative;
 import net.openhft.chronicle.core.annotation.UsedViaReflection;
-import net.openhft.chronicle.core.io.AbstractReferenceCounted;
-import net.openhft.chronicle.core.io.IORuntimeException;
-import net.openhft.chronicle.core.io.ReferenceOwner;
-import net.openhft.chronicle.core.io.UnsafeText;
+import net.openhft.chronicle.core.io.*;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -54,7 +54,10 @@
         extends AbstractReferenceCounted
         implements Bytes<U>,
         HasUncheckedRandomDataInput {
-    protected static final boolean DISABLE_THREAD_SAFETY = Jvm.getBoolean("disable.thread.safety", false);
+    protected static final boolean DISABLE_SINGLE_THREADED_CHECK =
+            Jvm.getBoolean("disable.single.threaded.check",
+                    Jvm.getBoolean("disable.thread.safety", false));
+    protected static final boolean DISABLE_THREAD_SAFETY = DISABLE_SINGLE_THREADED_CHECK;
```
nicktindall: Are we going to phase out `disable.thread.safety` in the future? should we warn if we detect people using it?

nicktindall: Also I wonder if we should expose the value from AbstractCloseable and defer to that? then we could just have it detected/set in one place?

nicktindall: e.g. `AbstractCloseable.areThreadSafetyChecksDisabled()`

nicktindall: Or even `Jvm.areThreadSafetyChecksDisabled()` which AbstractCloseable and this both defer to?

peter-lawrey: Jvm or SingleThreadedChecked methods would be best.

JerryShea:
See also https://github.com/OpenHFT/Chronicle-Core/pull/426 and [the build](https://teamcity.chronicle.software/buildConfiguration/Chronicle_BuildAll_Release/697073?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true)

nicktindall (approved):

PR #394, opened 2022-06-21
de-flake Windows tests

JerryShea:


src/test/java/net/openhft/chronicle/bytes/MappedMemoryTest.java
```diff
@@ -229,6 +205,9 @@ public void mappedMemoryTestSingle()
                 }
             }
         } finally {
+            // don't know why this test consistently shows this
+            if (OS.isWindows())
+                ignoreException("Unable to delete");
```
nicktindall, line 209: I remember reading that mapped segments aren't released until the object is garbage collected, so I assume even if you clean up resources it can still have the underlying file open? (see below from [map javadoc](https://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileChannel.html#map(java.nio.channels.FileChannel.MapMode,%20long,%20long)))
> The buffer and the mapping that it represents will remain valid until the buffer itself is garbage-collected.

nicktindall (approved):

src/test/java/net/openhft/chronicle/bytes/MappedMemoryTest.java
JerryShea: Yep - agree

PR #395, opened 2022-06-23
Demarshalling shouldn't leave an object bound to a thread, https://gi…

peter-lawrey:
…thub.com/OpenHFT/Chronicle-Wire/issues/520


alamar (approved):
Not sure about the class being moved

peter-lawrey:
That is the formatter. When we do x.24 I want to reformat all the non-generated code.

PR #407, opened 2022-07-20
Bytes/issue/406

peter-lawrey:


src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -1234,12 +1242,14 @@ public String toString() {
     @Override
     public void nativeRead(@NonNegative long position, long address, @NonNegative long size)
             throws BufferUnderflowException, IllegalStateException {
+        ensureCapacity(position+size);
```
JerryShea, line 1245: minor, but please `s/position+size/position + size/g`

src/main/java/net/openhft/chronicle/bytes/BytesStore.java
```diff
@@ -55,48 +54,6 @@
...
-    @SuppressWarnings("deprecation")
-    @Override
-    default boolean compareAndSwapDouble(@NonNegative long offset, double expected, double value)
-            throws BufferOverflowException, IllegalStateException {
-        return compareAndSwapLong(offset, Double.doubleToRawLongBits(expected), Double.doubleToRawLongBits(value));
-    }
-
-    @SuppressWarnings("deprecation")
-    @Override
-    default int addAndGetInt(@NonNegative long offset, int adding)
```
JerryShea: have these just moved? Do we have to move things around in files? It makes archaeology so much harder

src/main/java/net/openhft/chronicle/bytes/internal/CommonMappedBytes.java
```diff
@@ -431,7 +431,7 @@ void performClose() {
 
     @Override
     public boolean isClosed() {
-        return closeable.isClosed();
+        return closeable != null && closeable.isClosed();
```
JerryShea: how can closeable be null when it is set `private final`?

JerryShea (changes requested):
Can we undo the changes that move methods in BytesStore, as this will prevent us backporting any fixes to BytesStore going forward

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -3089,7 +3089,8 @@ public static void appendTimeMillis(@NotNull ByteStringAppender b, long timeInMS
     public static boolean equalBytesAny(@NotNull BytesStore b1, @NotNull BytesStore b2, @NonNegative long readRemaining)
             throws BufferUnderflowException, IllegalStateException {
 
-        if (b1.readRemaining() < readRemaining)
+        if (b1.readRemaining() < readRemaining
```
RobAustin: could use Math.min() .. but this is ok as well

RobAustin (approved):

src/main/java/net/openhft/chronicle/bytes/internal/CommonMappedBytes.java
peter-lawrey: This can be called when tracing it turned on before being set.

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
peter-lawrey: Done

src/main/java/net/openhft/chronicle/bytes/BytesStore.java
peter-lawrey: I have moved them back.

Issue #410, opened 2022-07-29
Bytes.wrapForWrite(ByteBuffer).write() corrupts heap


alamar:
I am attaching a test case based on Network SSL test I was looking into.

I suspect this is due to

            BytesStore<?, ByteBuffer> bs = BytesStore.wrap(byteBuffer);
            try {
                Bytes<ByteBuffer> bbb = bs.bytesForWrite();
                bbb.writePosition(byteBuffer.position());
                bbb.writeLimit(byteBuffer.limit());
                return bbb;
            } finally {
                bs.release(INIT); // what happens to bbb at this point?
            }

alamar:
#411

hft-team-city:
Released in [Chronicle-Bytes-2.23ea26](https://github.com/OpenHFT/Chronicle-Bytes/releases/chronicle-bytes-2.23ea26), [BOM-2.23ea112](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.23ea112)

PR #411, opened 2022-07-29
Document how to avoid heap corruption #410

alamar:
This test might need to be run a couple times, then it would corrupt heap for good, producing a multitude of malloc/free errors, successful JVM crash and even hanging JVM crash.


alamar:
@peter-lawrey Can you please take a look and provide recommendations on this fix?

alamar:
See also OpenHFT/Chronicle-Network#208

peter-lawrey (approved):

PR #413, opened 2022-08-02
undo <= change from previous commit, fix off-by-one errors in ChunkedMappedBytes and add test. Closes #412

### LINKED ISSUE #412: MappedBytesStore.inside() off by one error
This undoes the `<=` change from https://github.com/OpenHFT/Chronicle-Bytes/pull/243, beefs up testing and fixes some other off-by-ones

JerryShea:


src/test/java/net/openhft/chronicle/bytes/MappedBytesEdgeTest.java
```diff
@@ -0,0 +1,117 @@
...
+                bytes.writePosition(chunk + overlap - size);
+                doit.accept(bytes);
+                assertEquals(0, bytes.bytesStore().start());
+                bytes.writePosition(chunk + overlap);
+                doit.accept(bytes);
+                assertEquals(chunk, bytes.bytesStore().start());
+                bytes.writePosition(chunk + overlap + size);
+                doit.accept(bytes);
+                assertEquals(chunk, bytes.bytesStore().start());
+                // now try and write over the join - this should throw
```
JerryShea: TODO: remove and address later

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedBytes.java
```diff
@@ -379,25 +379,6 @@ public Bytes<Void> clear()
         return this;
     }
 
-    @NotNull
-    @Override
-    public Bytes<Void> writeByte(final byte i8)
```
JerryShea, line 384: I believe the superclass method does what we need

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedBytes.java
```diff
@@ -442,28 +423,13 @@ void rawCopy(@NonNegative final long length, final long fromAddress)
         uncheckedWritePosition(writePosition() + length);
     }
 
-    @Override
-    @NotNull
-    public Bytes<Void> writeOrderedInt(@NonNegative long offset, int i)
```
JerryShea, line 447: I believe the superclass method does what we need

src/main/java/net/openhft/chronicle/bytes/MappedBytesStore.java
```diff
@@ -111,8 +111,8 @@ public boolean inside(@NonNegative long offset) {
 
     @Override
     public boolean inside(@NonNegative long offset, @NonNegative long buffer) {
-        // this is correct that it uses the maximumLimit, yes it is different than the method above.
-        return start <= offset && offset + buffer <= limit;
+        // yes this is different than the method above
+        return start <= offset && offset + buffer < limit;
```
JerryShea, line 115: undid the <= from previous change

src/main/java/net/openhft/chronicle/bytes/BytesStore.java
```diff
@@ -327,6 +328,11 @@ default boolean inside(@NonNegative long offset) {
 
     /**
      * Returns if a number of bytes starting from an offset are inside this ByteStore limits.
+     * If you are going to write n bytes starting at offset, you need to call this method with buffer=n-1,
+     * as the 1st byte is written at offset, and the last at offset+n-1
+     * <p>
+     * Use this test to determine if an offset is considered safe to write to. Note that it checks we are
+     * inside the BytesStore limits *including* the overlap
```
nicktindall, line 335: I think the reason these errors exist is because this API is counter-intuitive. I reckon we should schedule a breaking change where the parameters for this method are interpreted as offset and length, meaning to check for a single byte you'd use the above method and to check for N = 2 or more bytes you'd call `inside(offset, N)`

nicktindall: Although I appreciate that is an atrocious API change to make.

nicktindall (approved):
I'll approve the changes, but I think we should seriously consider changing `inside(long, long)` to interpret the second parameter as length, because the current interpretation is VERY CLUMSY and the changes required to all the places it's called demonstrate that.
I wonder if in fact the bug is in that interpretation, and not at all the call sites. If the interpretation stays as "length after the offset byte" we will no doubt see more such bugs introduced in the future.

src/main/java/net/openhft/chronicle/bytes/BytesStore.java
peter-lawrey: It would be better to create a new method name and deprecate the old one.

peter-lawrey (approved):

src/main/java/net/openhft/chronicle/bytes/BytesStore.java
nicktindall: @JerryShea can we do this please? I really can see this being the source of a lot of subtle bugs going forward

PR #423, opened 2022-08-05
Fixed wrong writePosition being used for EmbeddedBytes

### LINKED ISSUE #422: Embedded bytes can't be cleared after TriviallyCopyable unmarshalling
The bug is that `AbstractBytes#writePosition` used as field somewhere, while it's overridden in `EmbeddedBytes`:
```
    @Override
    public @NonNegative long writePosition() {
        return bytesStore.readUnsignedByte(bytesStore.start() - 1);
    }
```
So we end up using wrong `writePosition` in some cases. 
Test:
```
    @Test
    public void testClear() {
        ClassAliasPool.CLASS_ALIASES.addAlias(EBM.class);
        EBM e1 = new EBM();
        e1.a.append("a12345678");
        Bytes<?> bytes = Bytes.allocateElasticOnHeap();
        e1.writeMarshallable(bytes);

        EBM e2 = new EBM();
        e2.readMarshallable(bytes);
        e2.a.clear();
        e2.a.append("b0000000");

        Assert.assertEquals("b0000000", e2.a.toString());

        bytes.releaseLast();
    }
```
Currently, `clear()` does nothing - because of this fast path check:
```
    public Bytes<U> clear()
        final long start = start();
        final long capacity = capacity();
        if (readPosition == start && writePosition == start && writeLimit == capacity) 
            return this;
```

glukos:
See also https://github.com/OpenHFT/Chronicle-Wire/pull/549 - reproducing test

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
@@ -64,7 +64,8 @@ public abstract class AbstractBytes<U>
     @NotNull
     protected BytesStore<Bytes<U>, U> bytesStore;
     protected long readPosition;
-    private long writePosition; // May be overridden in subclasses - use writePosition() getter to access it
+    // Make private in x.24 - can be overridden in subclasses, should be accessed via writePosition() getter
+    protected long writePosition;
     protected long writeLimit;
     protected boolean isPresent;
     private int lastDecimalPlaces = 0;
--- a/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
@@ -64,7 +64,7 @@ public abstract class AbstractBytes<U>
     @NotNull
     protected BytesStore<Bytes<U>, U> bytesStore;
     protected long readPosition;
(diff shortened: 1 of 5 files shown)


JerryShea (approved):

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -133,7 +133,7 @@ public Bytes<U> clear()
             throws IllegalStateException {
         final long start = start();
         final long capacity = capacity();
-        if (readPosition == start && writePosition == start && writeLimit == capacity)
+        if (readPosition == start && writePosition() == start && writeLimit == capacity)
```
nicktindall, line 137: Good find, I wonder if it's feasible to/worth making `writePosition` private to reduce the number of places a bug like this can be reintroduced going forward?

nicktindall (approved):

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
glukos: Good point.
It's a compatibility break to make this change right away - I've left a comment to do it in x.24

Issue #426, opened 2022-08-24
Bytes.read(byte[]) fails if Bytes constructed with a BytesStore that is a Bytes


alamar:
At `net.openhft.chronicle.bytes.RandomDataInput#read`, commit #414 has introduced a change which triggers the following exception:

```
Exception in thread "Thread-2" java.lang.ArithmeticException: Unsigned Int 31-bit -28 out of range
	at net.openhft.chronicle.core.Maths.toUInt31(Maths.java:659)
	at net.openhft.chronicle.bytes.RandomDataInput.read(RandomDataInput.java:664)
	at net.openhft.chronicle.bytes.AbstractBytes.read(AbstractBytes.java:1124)
	at net.openhft.chronicle.bytes.StreamingDataInput.read(StreamingDataInput.java:384)
	at net.openhft.chronicle.bytes.VanillaBytes.read(VanillaBytes.java:702)
	at software.chronicle.enterprise.queue.demo.example13.QueueMultiThreadedJLBHBenchmark$Datum.readMarshallable(QueueMultiThreadedJLBHBenchmark.java:133)
```

In this case, 

```
            int len = Math.min(length, Maths.toUInt31(readLimit() - offsetInRDI));
//                                                              0 - 28
```

JerryShea:
What made this happen was some code in a demo that constructed a `Bytes` with a `BytesStore` that is a `Bytes` e.g. `new OnHeapBytes(Bytes.allocateElasticOnHeap(capacity), true)`.

See #427 for some tests that fail if you do this.

Previously in `net.openhft.chronicle.bytes.RandomDataInput#read` there was a cast to `int` which gave a -ve number and the actual read would silently fail. Now we use `Maths.toUInt31` which blows up if you give it a -ve number which arguably is better as it is demonstrating a problem further up the tree (and has enabled us to find this).

JerryShea:
Closed by https://github.com/OpenHFT/Chronicle-Bytes/pull/428

PR #427, opened 2022-08-25
Issue/426 read out of range

JerryShea:
DO NOT MERGE

This cleans up a few things and adds some tests but does not fix the problem exposed by `Allocator.HEAP_BYTES_BYTESSTORE` - see #426

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/internal/HeapBytesStore.java
+++ b/src/main/java/net/openhft/chronicle/bytes/internal/HeapBytesStore.java
@@ -24,6 +24,7 @@ import net.openhft.chronicle.core.Jvm;
 import net.openhft.chronicle.core.UnsafeMemory;
 import net.openhft.chronicle.core.annotation.NonNegative;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.nio.BufferOverflowException;
 import java.nio.BufferUnderflowException;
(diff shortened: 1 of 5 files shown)


src/main/java/net/openhft/chronicle/bytes/internal/NativeBytesStore.java
```diff
@@ -839,8 +839,7 @@ public boolean sharedMemory() {
     @Override
     public long read(@NonNegative long offsetInRDI, byte[] bytes, @NonNegative int offset, @NonNegative int length) {
         requireNonNull(bytes);
-        // length is non-neg and an int
-        int len = (int) Math.min(length, readLimit() - offsetInRDI);
+        int len = Math.min(length, Maths.toUInt31(readLimit() - offsetInRDI));
```
JerryShea, line 842: this reverts https://github.com/OpenHFT/Chronicle-Bytes/commit/b0324d677cdad0d74771c1f5fc871044e8be8add

We are better having this blow up if `readLimit() - offsetInRDI` is negative, rather than silently fail

src/test/java/net/openhft/chronicle/bytes/StreamingDataInputTest.java
```diff
@@ -45,6 +45,17 @@ public static Object[] params() {
 
     @Test
     public void read() {
+        Bytes<?> b = bytesType.createBuffer();
```
JerryShea: test for previously uncovered code

src/main/java/net/openhft/chronicle/bytes/internal/HeapBytesStore.java
```diff
@@ -139,7 +140,6 @@ protected void performRelease() {
         return capacity;
     }
 
-    @NotNull
```
nicktindall, line 142: I'm still unclear on how these annotations are interpreted, do we need to add a @Nullable annotation here or do we assume anything not @NotNull is @Nullable? (and if that's the case, why have the annotation on the constructor arg?)

nicktindall (approved):
Always supportive of not silently failing :)

PR #433, opened 2022-08-31
fix for bounds check issue when using PointsByteStore, Fixes #432

### LINKED ISSUE #432: PointerBytesStore bounds check failing when attempting to read from direct bytes

vach-chronicle:


src/main/java/net/openhft/chronicle/bytes/PointerBytesStore.java
```diff
@@ -33,14 +36,18 @@ public PointerBytesStore() {
     public void set(long address, @NonNegative long capacity) {
         setAddress(address);
         this.limit = maximumLimit = capacity;
+        if (capacity == Bytes.MAX_CAPACITY)
+            Jvm.warn().on(getClass(), "the provided capacity of underlying looks like it may have come " +
```
JerryShea, line 40: maybe say that we won't support this in e.g. x.25 and will throw an exception?

vach-chronicle:
Problem is we have no 100% way to know that its elastic,
if the provided capacity matches then we have a hint... otherwise no way
that i know of.

Vachagan Balayan
+16572318413
chronicle.software


Chronicle Software Ltd
Registered in England with Company Number 8381496
at 82 St John Street,
<https://maps.google.com/?q=82+St+John+Street,+++Clerkenwell,+London+EC1M+4JN,+UK&entry=gmail&source=g>Clerkenwell,
London EC1M 4JN, UK
<https://maps.google.com/?q=82+St+John+Street,+++Clerkenwell,+London+EC1M+4JN,+UK&entry=gmail&source=g>

src/main/java/net/openhft/chronicle/bytes/PointerBytesStore.java
vach-chronicle: The problem is i can't detect if it's elastic with 100% certainty
all we can do is check the capacity and warn them

JerryShea: fair enough :)

JerryShea (approved):
LGTM but I would wait for PL/RA review

src/main/java/net/openhft/chronicle/bytes/PointerBytesStore.java
```diff
@@ -18,11 +18,14 @@
 package net.openhft.chronicle.bytes;
 
 import net.openhft.chronicle.bytes.internal.NativeBytesStore;
+import net.openhft.chronicle.core.Jvm;
 import net.openhft.chronicle.core.annotation.NonNegative;
 import org.jetbrains.annotations.NotNull;
 
 /**
  * A BytesStore which can point to arbitrary memory.
+ * Used to act as a view of Bytes over an area of memory.
```
RobAustin: suggestion :  Acts as a view of Bytes over an area of memory.

RobAustin (approved):

PR #441, opened 2022-09-12
Observe append8bit (start, end) API

### LINKED ISSUE #440: null byte in JSON representation
There seems to be some cases where JSONWire adds NULL byte(s) to the JSON representation. So far I have only seen this when setting` use8bit`  to true. Here is a minimal example that shows that behavior:

```java
    @Test
    void example() {
        final Bytes<byte[]> bytes = Bytes.allocateElasticOnHeap(1024);
        final JSONWire wire = new JSONWire(bytes, true);

        final Foo foo = new Foo();

        wire.getValueOut().marshallable(foo);

        System.out.println(bytes.toHexString());
    }

    private static class Foo extends SelfDescribingMarshallable {
        final Bar a = new Bar();

        private static class Bar extends SelfDescribingMarshallable {
            String b = "c";
        }
    }
```

The output is below and contains a NULL byte before the closing "}":
```
00000000 22 61 22 3a 7b 22 62 22  3a 22 63 22 00 7d       "a":{"b" :"c"·} 
```

This bit of code in ByteStringAppender looks suspicious:
```java
    default B append8bit(@NotNull CharSequence cs, int start, int end)
            throws IllegalArgumentException, BufferOverflowException, BufferUnderflowException, IndexOutOfBoundsException, IllegalStateException {
        if (cs instanceof BytesStore) {
            return write((BytesStore) cs, (long) start, end);
        }
...
```

`start` and `end` are being passed to `write` which expects an offset and length: `default S write(@NotNull BytesStore bytes, long offset, long length)`.

alamar:
fixes #440

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/ByteStringAppender.java
+++ b/src/main/java/net/openhft/chronicle/bytes/ByteStringAppender.java
@@ -337,8 +337,9 @@ public interface ByteStringAppender<B extends ByteStringAppender<B>> extends Str
      */
     default B append8bit(@NotNull CharSequence cs, @NonNegative int start, @NonNegative int end)
             throws IllegalArgumentException, BufferOverflowException, BufferUnderflowException, IndexOutOfBoundsException, IllegalStateException {
+        assert end >= start : "end=" + end + ",start=" + start;
         if (cs instanceof BytesStore) {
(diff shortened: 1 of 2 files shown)


RobAustin (approved):

alamar:
See also OpenHFT/Chronicle-Wire#572

Issue #443, opened 2022-09-22
The CanonicalPath should not be unique when run with different classLoaders in the same JVM


RobAustin:
When running in different classloaders,  the following issue was found

![image](https://user-images.githubusercontent.com/6506056/191758556-b3c9de20-1b54-4837-87a7-34098e8b2292.png)

hft-team-city:
Released in [Chronicle-Bytes-2.23.31](https://github.com/OpenHFT/Chronicle-Bytes/releases/chronicle-bytes-2.23.31), [BOM-2.23.126](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.23.126)

PR #444, opened 2022-09-22
The CanonicalPath should not be unique when run with different classLoaders in the same JVM

RobAustin:
See the [ #433](https://github.com/OpenHFT/Chronicle-Bytes/issues/443) issue for details

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/internal/CanonicalPathUtil.java
+++ b/src/main/java/net/openhft/chronicle/bytes/internal/CanonicalPathUtil.java
@@ -17,6 +17,7 @@
  */
 package net.openhft.chronicle.bytes.internal;
 
+import net.openhft.chronicle.core.Jvm;
 import net.openhft.chronicle.core.io.IORuntimeException;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/bytes/internal/CanonicalPathUtil.java
```diff
@@ -27,7 +28,7 @@
 public final class CanonicalPathUtil {
 
     // This is unlikely to collide with other internalized strings and is seldom the same across JVMs
-    private static final String PREFIX = CanonicalPathUtil.class.getName() + Math.abs(new SecureRandom().nextInt(1 << 30)) + "::";
```
JerryShea, line 30: I believe that this commit - https://github.com/OpenHFT/Chronicle-Bytes/commit/58d297f9ec11a55378da5a37238ade6c1f7b7647 - broke this behaviour and that we should revert to synchronizing on
```
file.getCanonicalPath().intern()
```
i.e. we should get rid of PREFIX altogether e.g. https://github.com/OpenHFT/Chronicle-Bytes/tree/feature/canonical-path

JerryShea: See also https://github.com/OpenHFT/Chronicle-Bytes/pull/445

JerryShea:
Closing as using changes from #445 (#447)

PR #453, opened 2022-09-29
Fix acquireNextByteStore0 when shifting backwards

### LINKED ISSUE #452: Acquiring next byte store may not change the state when shifting backwards
```
        if (adding > 0 && !bytesStore.inside(offset, checkSize0(adding - 1))) {
            acquireNextByteStore0(offset + adding - 1, false);
```
The logic above fails when `bytesStore.start()` is greater than `offset`, but the difference between start and `offset` is less than `adding`.
Real-life example is overwriting the header after we wrote a message.

glukos:


src/test/java/net/openhft/chronicle/bytes/MappedBytesTest.java
```diff
@@ -91,6 +92,33 @@ public void testMappedFileSafeLimitTooSmall()
...
+            bytesW.writeInt(7);
+            Assert.assertEquals(chunkSize * 2 + 4, bytesW.writePosition());
+
+            bytesW.writeInt(chunkSize * 2 - 2, 9);
+
+            bytesW.readPosition(chunkSize * 2 - 2);
+            Assert.assertEquals(9, bytesW.readInt());
+
+
+        } catch (Exception e) {
```
JerryShea: minor, but we probably don't need the catch block at all

JerryShea (approved):

nicktindall (approved):

PR #456, opened 2022-10-07
Issue/454 backport fix

JerryShea:


nicktindall (approved):

alamar (approved):
Approve, but we should make sure the test is also in later branches.

Issue #454, opened 2022-10-06
Boundary underflow in chunked mapped file in multiple operations near chunk boundary


alamar:
```
Exception in thread "main" net.openhft.chronicle.bytes.util.DecoratedBufferUnderflowException: Acquired the next BytesStore, but still not room to add 2 when realCapacity 15728640
	at net.openhft.chronicle.bytes.internal.ChunkedMappedBytes.writeCheckOffset(ChunkedMappedBytes.java:266)
	at net.openhft.chronicle.bytes.AbstractBytes.writeOffsetPositionMoved(AbstractBytes.java:1023)
	at net.openhft.chronicle.bytes.AbstractBytes.writeOffsetPositionMoved(AbstractBytes.java:1016)
	at net.openhft.chronicle.bytes.AbstractBytes.writeShort(AbstractBytes.java:1043)
	at org.example.Main.main(Main.java:38)
```

alamar:
Broken by commit
```
commit e6598e3723e04248670da493761326b36c45617a (HEAD)
Author: Jerry Shea <jerry.shea@chronicle.software>
Date:   Wed Aug 3 11:01:41 2022 +1000

    undo <= change from previous commit and add test. Add MappedBytesEdgeTest and address off-by-ones. Closes #412. Fix ChunkedMappedBytes.ensureCapacity. Make acquireNextByteStore0 use 3 args. Closes #417. Clean up use of API

```
@JerryShea can you please take a look?

alamar:
https://github.com/OpenHFT/Chronicle-Bytes/pull/455 a test for this issue

alamar:
`acquireNextByteStore0()` in `ChunkedMappedBytes:272` now fails to pullback to previous chunk when we need to write message starting in it.

alamar:
See #452

alamar:
Supposedly fixed in develop by b305af85c6f3fe0ebe60df7c58b9477f5b04ae7c

hft-team-city:
Released in [Chronicle-Bytes-2.23.33](https://github.com/OpenHFT/Chronicle-Bytes/releases/chronicle-bytes-2.23.33), [BOM-2.23.134](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.23.134)

hft-team-city:
[...]

PR #457, opened 2022-10-07
A test for issue #454

alamar:
It seems that we have an issue with ChunkedMappedBytes not having a BytesStore - let's discuss.


alamar:
Please review test addition @glukos @JerryShea

alamar:
@glukos @JerryShea please review

PR #458, opened 2022-10-07
A test for issue #454 (fails)

alamar:
The issue rears its ugly head - we have a regression where the user scenario no longer works since bytesForWrite() now returns a view of empty heap bytes store (MappedChunkedBytes having no store of its own).

@JerryShea can you please take a look?

See also #457 and #456

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
+++ b/src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
@@ -3404,7 +3404,8 @@ enum BytesInternal {
     }
 
     public static <B extends BytesStore<B, U>, U> BytesStore<B, U> failIfBytesOnBytes(BytesStore<B, U> bytesStore) {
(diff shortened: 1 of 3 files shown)


src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -3404,7 +3404,8 @@ static class DateCache {
     }
 
     public static <B extends BytesStore<B, U>, U> BytesStore<B, U> failIfBytesOnBytes(BytesStore<B, U> bytesStore) {
-        if (bytesStore instanceof Bytes) {
+        // MappedBytes don't have a backing BytesStore so we have to allow them to be used this way
```
alamar, line 3407: Can you please create a separate issue to remove this check and introduce MappedBytesStore somewhere in the future?

alamar:
Can't approve my own PR, please approve and merge.

JerryShea (approved):

PR #460, opened 2022-10-18
warn if no PosixAPI available (rather than NPE)

JerryShea:


peter-lawrey:
The warning should be a once off

JerryShea:
> The warning should be a once off

well then maybe we don't warn at all and rely on https://github.com/OpenHFT/Posix/pull/21 instead - what do you think?

peter-lawrey:
The psox should return a dummy implementation that either does nothing or always returns an appropriate error

JerryShea:
Replaced by https://github.com/OpenHFT/Posix/pull/21/

Issue #469, opened 2022-11-09
vectorisedMismatch falsely equates single character strings


nicktindall:
When `BytesStore#contentEquals` is performed using vectorised mismatch (Java 11+ with `--illegal-access=permit --add-exports java.base/jdk.internal.ref=ALL-UNNAMED --add-exports java.base/jdk.internal.util=ALL-UNNAMED`) and the two strings are single character, it always returns true regardless of if they are the same character.

nicktindall:
It's breaking the following tests:

https://teamcity.chronicle.software/test/-7506061407356869674?currentProjectId=Chronicle_BuildAll_Checks

https://teamcity.chronicle.software/test/-878476738362959368?currentProjectId=Chronicle_BuildAll_Checks

It's only popped up now because I bumped the parent POM which added the requisite JVM args to enable this behaviour

hft-team-city:
Released in [Chronicle-Bytes-2.23.33](https://github.com/OpenHFT/Chronicle-Bytes/releases/chronicle-bytes-2.23.33), [BOM-2.23.134](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.23.134)

hft-team-city:
Released in [Chronicle-Bytes-2.24ea1](https://github.com/OpenHFT/Chronicle-Bytes/releases/chronicle-bytes-2.24ea1), [BOM-2.24ea9](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.24ea9)

hft-team-city:
Released in [Chronicle-Bytes-2.22.29](https://github.com/OpenHFT/Chronicle-Bytes/releases/chronicle-bytes-2.22.29), [BOM-2.22.172](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.22.172)

PR #470, opened 2022-11-09
Reproduction for #469

nicktindall:
Whoever ends up fixing this can do it on this branch then merge


src/test/java/net/openhft/chronicle/bytes/BytesTest.java
```diff
@@ -866,6 +866,18 @@ public void testToStringDoesNotChange() {
         }
     }
 
+    @Test
+    public void contentEqualsSingleChar() {
+        @NotNull Bytes<?> a = alloc1.elasticBytes(16);
```
JerryShea, line 871: TODO: we really need some proper generational style tests for this

JerryShea (approved):

PR #477, opened 2022-11-20
Temp/instanceof

JerryShea:


src/main/java/net/openhft/chronicle/bytes/ByteStringAppender.java
```diff
@@ -339,7 +339,7 @@ default B append8bit(@NotNull CharSequence cs, @NonNegative int start, @NonNegat
             throws IllegalArgumentException, BufferOverflowException, BufferUnderflowException, IndexOutOfBoundsException, IllegalStateException {
         assert end >= start : "end=" + end + ",start=" + start;
         if (cs instanceof BytesStore) {
-            return write((BytesStore) cs, (long) start, end - start);
+            return write((BytesStore) cs, start, (long)end - start);
```
JerryShea, line 342: this fixes a Sonar warning

Issue #482, opened 2023-01-04
When finished warmUp() return the Jvm loggers back to what they were set to 


RobAustin:
When finished warmUp() return the Jvm loggers back to what they were set to, rather than assuming that they were set to SLF loggers and resetting back the Slf loggers

JerryShea:
Closing as I believe done

PR #483, opened 2023-01-04
When finished warmUp() return the Jvm loggers back

RobAustin:
see issue #482


src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedFile.java
```diff
@@ -80,8 +82,14 @@ public ChunkedMappedFile(@NotNull final File file,
 
     public static void warmup() {
         final List<IOException> errorsDuringWarmup = new ArrayList<>();
+
```
maxim-ponomarev, line 85:
```suggestion
```

maxim-ponomarev (approved):

PR #487, opened 2023-02-08
Implement ReentrantThreadScopedFileLock

nicktindall:
An approach to fixing https://github.com/OpenHFT/Chronicle-Queue/issues/1291 (see also #488)

This approach attempts to prevent threads from the same VM taking overlapping file locks. Here if two threads attempt to lock the file using `ReentrantThreadScopedLock` the second thread will block until the first thread releases the lock.

This could be dangerous, one example is in `resizeRafIfTooSmall`; there used to be two locks acquired to do a resizeRaf..:

a. The file lock
b. The monitor on the canonical path string

In one execution path in Chronicle-Queue we acquire (a), then (b), then (a) again. This is the cause of https://github.com/OpenHFT/Chronicle-Queue/issues/1291. If two different processes did that it wouldn't introduce a deadlock because the thread holding (b) would throw an `OverlappingFileLockException` when it tried to acquire (a), but with `ReentrantThreadScopedLock` it could result in a deadlock if thread 1 holds (a) and is waiting on (b) while thread 2 holds (b) and is waiting on (a).

I removed the synchronization on the monitor to remove that possiblity. The `ReentrantThreadScopedLock` should provide the same level of safety.

## Limitation
A limitation to this approach is the set of current locks is kept in a static variable in `ReentrantThreadScopedLock` so it won't work for two threads that have loaded `ReentrantThreadScopedLock` from different class loaders. In this scenario, OverlappingFileLockExceptions will still be thrown.

It would have been nicer to use the canonical path string to mutex across class loaders, but there's no way for Java 9+ to do a tryLock on a mutex ([Unsafe.tryMonitorEnter is deprecated in Java 8](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8069302))

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedFile.java
+++ b/src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedFile.java
@@ -25,7 +25,6 @@ import net.openhft.chronicle.core.annotation.NonNegative;
 import net.openhft.chronicle.core.io.IORuntimeException;
 import net.openhft.chronicle.core.io.ReferenceOwner;
 import net.openhft.chronicle.core.onoes.ExceptionHandler;
-import net.openhft.chronicle.core.onoes.Slf4jExceptionHandler;
 import net.openhft.chronicle.core.onoes.ThreadLocalisedExceptionHandler;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -242,7 +241,7 @@ public class ChunkedMappedFile extends MappedFile {
             return;
 
         // handle a possible race condition between processes.
-        try {
+        try (final ReentrantThreadScopedFileLock lock = ReentrantThreadScopedFileLock.lock(file(), fileChannel)) {
             // A single JVM cannot lock a distinct canonical file more than once.
 
             // We might have several MappedFile objects that maps to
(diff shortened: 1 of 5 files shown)


vach-chronicle (approved):

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedFile.java
```diff
@@ -242,7 +241,7 @@ private void resizeRafIfTooSmall(@NonNegative final int chunk)
             return;
 
         // handle a possible race condition between processes.
-        try {
+        try (final ReentrantThreadScopedFileLock lock = ReentrantThreadScopedFileLock.lock(file(), fileChannel)) {
```
nicktindall, line 244: If you've acquired this lock you know you're the only one on the machine with it.

src/main/java/net/openhft/chronicle/bytes/internal/SingleMappedFile.java
```diff
@@ -120,7 +120,7 @@ private void resizeRafIfTooSmall(@NonNegative final long minSize)
             return;
 
         // handle a possible race condition between processes.
-        try {
+        try (final ReentrantThreadScopedFileLock lock = ReentrantThreadScopedFileLock.lock(file(), fileChannel)) {
```
nicktindall, line 123: If you've acquired this lock you know you're the only one on the machine with it.

PR #495, opened 2023-03-15
Fix error is asDouble and parseDouble on the edge of precision. 

peter-lawrey:
Issue https://github.com/OpenHFT/Chronicle-Core/issues/484

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
@@ -234,22 +234,18 @@ public abstract class AbstractBytes<U>
     @Override
     public @NotNull AbstractBytes<U> append(double d)
             throws BufferOverflowException, IllegalStateException {
-        boolean fits = canWriteDirect(380);
-        double ad = Math.abs(d);
-        if (ad < 1e-18) {
-            append(Double.toString(d));
-            return this;
-        }
-        if (!fits) {
-            fits = 1e-6 <= ad && ad < 1e20 && canWriteDirect(24);
-        }
+        boolean fits = canWriteDirect(32);
         if (fits) {
             long address = addressForWrite(writePosition());
             long address2 = UnsafeText.appendDouble(address, d);
             writeSkip(address2 - address);
             return this;
+        } else {
+            Bytes<?> bytes = BytesInternal.acquireBytes();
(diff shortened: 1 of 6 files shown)


src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -234,22 +234,18 @@ public boolean compareAndSwapLong(@NonNegative long offset, long expected, long
...
-        boolean fits = canWriteDirect(380);
-        double ad = Math.abs(d);
-        if (ad < 1e-18) {
-            append(Double.toString(d));
-            return this;
-        }
-        if (!fits) {
-            fits = 1e-6 <= ad && ad < 1e20 && canWriteDirect(24);
-        }
+        boolean fits = canWriteDirect(32);
```
nicktindall, line 237: Is 32 sufficient? is it because we put exponents in for really large/small numbers?

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -2141,41 +2141,8 @@ public static void append(@NotNull StreamingDataOutput out, double d)
...
-            d = value * fives;
-
-        } else {
-            d = value;
-        }
-
-        double scalb = Math.scalb(d, exp - deci - scale2);
-        return negative ? -scalb : scalb;
+    public static double asDouble(long value, int exp, boolean negative, int deci) {
+        return Maths.asDouble(value, exp, negative, deci);
```
nicktindall: Could we just delete this and delegate directly to the `Maths.asDouble(..)` instead? it was private previously so should be fine to remove.

nicktindall: With a static import it should still be a minimal change

nicktindall (approved):
It'd be good to remove that asDouble delegation from BytesInternal but I'll leave that up to you.

nicktindall:
Also seems to be quite a few broken tests still remaining.

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
peter-lawrey: Yes

PR #497, opened 2023-03-20
misc perf tweaks

JerryShea:


src/main/java/net/openhft/chronicle/bytes/VanillaBytes.java
```diff
@@ -276,6 +276,16 @@ public boolean isEqual(@Nullable String other)
             return bytesStore.realCapacity();
         }
 
+        @Override
+        public long findByte(byte stopByte) throws IllegalStateException {
```
JerryShea, line 280: so we don't have to rely on default methods

nicktindall (approved):

PR #498, opened 2023-03-21
Fix offset in acquire, inside() method

### LINKED ISSUE #481: Errors writing around chunk boundaries in non-forward-sequential access patterns
The way we check if a range is in a chunk, and load chunks when it is not, is broken in a way that fails when writing around chunk boundaries in non-forward-sequential access patterns.

nicktindall:
Fixes #481 

Hide whitespace when reviewing, one file had an extra indent through the whole file that I removed

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
@@ -54,7 +54,6 @@ public abstract class AbstractBytes<U>
         HasUncheckedRandomDataInput {
     @Deprecated(/* remove in x.25 */)
     protected static final boolean DISABLE_THREAD_SAFETY = SingleThreadedChecked.DISABLE_SINGLE_THREADED_CHECK;
-    private static final byte[] EMPTY_ARRAY = new byte[0];
     private static final boolean BYTES_BOUNDS_UNCHECKED = Jvm.getBoolean("bytes.bounds.unchecked", false);
 
     // used for debugging
@@ -218,21 +217,21 @@ public abstract class AbstractBytes<U>
     @Override
     public boolean compareAndSwapInt(@NonNegative long offset, int expected, int value)
             throws BufferOverflowException, IllegalStateException {
-        writeCheckOffset(offset, 4);
+        writeCheckOffset(offset, Integer.BYTES);
         return bytesStore.compareAndSwapInt(offset, expected, value);
     }
 
     @Override
     public void testAndSetInt(@NonNegative long offset, int expected, int value)
             throws BufferOverflowException, IllegalStateException {
-        writeCheckOffset(offset, 4);
+        writeCheckOffset(offset, Integer.BYTES);
         bytesStore.testAndSetInt(offset, expected, value);
     }
 
     @Override
     public boolean compareAndSwapLong(@NonNegative long offset, long expected, long value)
             throws BufferOverflowException, IllegalStateException {
-        writeCheckOffset(offset, 8);
+        writeCheckOffset(offset, Long.BYTES);
         return bytesStore.compareAndSwapLong(offset, expected, value);
     }
 
@@ -410,7 +409,7 @@ public abstract class AbstractBytes<U>
     public int readUnsignedByte()
             throws IllegalStateException {
         try {
-            long offset = readOffsetPositionMoved(1);
+            long offset = readOffsetPositionMoved(Byte.BYTES);
             return bytesStore.readUnsignedByte(offset);
 
         } catch (BufferUnderflowException e) {
@@ -439,7 +438,7 @@ public abstract class AbstractBytes<U>
     public byte readByte()
             throws IllegalStateException {
         try {
-            long offset = readOffsetPositionMoved(1);
+            long offset = readOffsetPositionMoved(Byte.BYTES);
             return bytesStore.readByte(offset);
 
         } catch (BufferUnderflowException e) {
@@ -464,7 +463,7 @@ public abstract class AbstractBytes<U>
     public short readShort()
             throws BufferUnderflowException, IllegalStateException {
         try {
-            long offset = readOffsetPositionMoved(2);
+            long offset = readOffsetPositionMoved(Short.BYTES);
             return bytesStore.readShort(offset);
         } catch (BufferUnderflowException e) {
             if (lenient) {
@@ -478,7 +477,7 @@ public abstract class AbstractBytes<U>
     public int readInt()
             throws BufferUnderflowException, IllegalStateException {
         try {
-            long offset = readOffsetPositionMoved(4);
+            long offset = readOffsetPositionMoved(Integer.BYTES);
             return bytesStore.readInt(offset);
         } catch (BufferUnderflowException e) {
             if (lenient) {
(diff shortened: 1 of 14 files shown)


src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedBytes.java
```diff
@@ -260,13 +260,10 @@ protected void writeCheckOffset(final @NonNegative long offset, final @NonNegati
         if (offset + adding < start() || offset > mappedFile.capacity() - adding)
             throw writeBufferOverflowException0(offset);
         BytesStore bytesStore = this.bytesStore;
-        if (adding > 0 && !bytesStore.inside(offset, checkSize0(adding - 1))) {
-            if (bytesStore.start() > offset)
-                acquireNextByteStore0(offset, false);
-            else
-                acquireNextByteStore0(offset + adding - 1, false);
+        if (adding > 0 && !bytesStore.inside(offset, checkSize0(adding))) {
+            acquireNextByteStore0(offset, false);
```
nicktindall, line 264: The old else is the cause of https://github.com/OpenHFT/Chronicle-Bytes/issues/452, this bug is repeated throughout the bytes implementations. The basic error is

```
if (!bytesStore.inside(offset, length)) {
     bytesStore.acquire(offset + length);
}
```
This works most of the time when you're writing forwards sequentially because if what you're writing won't fit in currentStore + offset, it'll usually fit in the store that gets acquired (nextStore) because that will start at currentStore + 1

The error shows up in the forward direction when you jump ahead a cycle (see what I added to the test)

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedBytes.java
```diff
@@ -452,8 +449,8 @@ public short readVolatileShort(@NonNegative long offset)
         throwExceptionIfClosed();
 
         BytesStore bytesStore = this.bytesStore;
-        if (!bytesStore.inside(offset, 1)) {
-            bytesStore = acquireNextByteStore0(offset + 1, false);
+        if (!bytesStore.inside(offset, 2)) {
+            bytesStore = acquireNextByteStore0(offset, false);
```
nicktindall, line 456: example of the bug

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedBytes.java
```diff
@@ -464,8 +461,8 @@ public int readVolatileInt(@NonNegative long offset)
         throwExceptionIfClosed();
 
         BytesStore bytesStore = this.bytesStore;
-        if (!bytesStore.inside(offset, 3)) {
-            bytesStore = acquireNextByteStore0(offset + 3, false);
+        if (!bytesStore.inside(offset, 4)) {
+            bytesStore = acquireNextByteStore0(offset, false);
```
nicktindall, line 468: and again

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedBytes.java
```diff
@@ -476,8 +473,8 @@ public long readVolatileLong(@NonNegative long offset)
         throwExceptionIfClosed();
 
         BytesStore bytesStore = this.bytesStore;
-        if (!bytesStore.inside(offset, 7)) {
-            bytesStore = acquireNextByteStore0(offset + 7, false);
+        if (!bytesStore.inside(offset, 8)) {
+            bytesStore = acquireNextByteStore0(offset, false);
```
nicktindall, line 480: and again

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedBytes.java
```diff
@@ -516,8 +513,8 @@ public int peekVolatileInt()
             throws IllegalStateException {
 
         BytesStore bytesStore = this.bytesStore;
-        if (!bytesStore.inside(readPosition, 3)) {
-            bytesStore = acquireNextByteStore0(readPosition + 3, true);
+        if (!bytesStore.inside(readPosition, 4)) {
+            bytesStore = acquireNextByteStore0(readPosition, true);
```
nicktindall, line 520: and again

src/main/java/net/openhft/chronicle/bytes/internal/NoBytesStore.java
```diff
@@ -309,8 +309,8 @@ public boolean inside(@NonNegative long offset) {
     }
 
     @Override
-    public boolean inside(@NonNegative long offset, @NonNegative long buffer) {
-        return false;
+    public boolean inside(@NonNegative long offset, @NonNegative long bufferSize) {
+        return offset == 0 && bufferSize == 0;
```
nicktindall, line 313: an array of zero length at position zero is inside the empty store

src/test/java/net/openhft/chronicle/bytes/MappedBytesEdgeTest.java
```diff
@@ -45,80 +48,93 @@ public MappedBytesEdgeTest(int size, ReadWrite rw, Consumer<Bytes<?>> doit) {
...
-                    // now try and write over the end of the chunk
-                    bytes.writePosition(chunk - 1);
-                    doit.accept(bytes);
-                    // and end of chunk plus offset
-                    bytes.writePosition(chunk + overlap - 1);
-                    doit.accept(bytes);
+                    // go back to the first store
+                    checkWritePosition(bytes, CHUNK_SIZE - 1, 0);
+                    // write to just before the start of the third bytes store (i.e. not in the 1st plus overlap or 2nd plus overlap)
+                    checkWritePosition(bytes, 3 * CHUNK_SIZE - size + 1, 2 * CHUNK_SIZE);
```
nicktindall: This reproduces the bug. If you pop the last commit (with the fix in it) off, you will see it blow.

src/main/java/net/openhft/chronicle/bytes/BytesStore.java
```diff
@@ -353,18 +353,13 @@ default boolean inside(@NonNegative long offset) {
 
     /**
      * Returns if a number of bytes starting from an offset are inside this ByteStore limits.
-     * If you are going to write n bytes starting at offset, you need to call this method with buffer=n-1,
-     * as the 1st byte is written at offset, and the last at offset+n-1
-     * <p>
-     * Use this test to determine if an offset is considered safe to write to. Note that it checks we are
-     * inside the BytesStore limits *including* the overlap
```
nicktindall, line 360: I still feel strongly that we make this intuitive rather than writing a convoluted comment as to why it's not and how to use it "correctly".

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedBytes.java
nicktindall: e.g.

for a mapped file with chunksize 100, overlap 20

the sequential case
```java
writeLong(50, some-long);   // positioned in the first chunk
writeLong(119, some-long);
  -> if (!bytesStore.inside(119, 8)) {     // false
          bytesStore.acquire(127);         // loads second store, which happens to contain 119->127, so OK
```

the non-sequential case (broken)
```java
writeLong(50, some-long);   // positioned in the first chunk
writeLong(199, some-long);
  -> if (!bytesStore.inside(199, 8)) {     // false
          bytesStore.acquire(207);         // loads third store, which does not contain 199->207, so FAIL
```

the backwards traversal case (broken)
```java
writeLong(150, some-long);   // positioned in the second chunk
writeLong(99, some-long);
  -> if (!bytesStore.inside(99, 8)) {     // false
          bytesStore.acquire(107);         // loads second store, which does not contain 99->107, so FAIL
```

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -213,21 +212,21 @@ public long start() {
     @Override
     public boolean compareAndSwapInt(@NonNegative long offset, int expected, int value)
             throws BufferOverflowException, IllegalStateException {
-        writeCheckOffset(offset, 4);
+        writeCheckOffset(offset, Integer.BYTES);
```
JerryShea, line 220: nice

src/main/java/net/openhft/chronicle/bytes/internal/ChunkedMappedBytes.java
peter-lawrey: Agreed, I saw that on Monday.

JerryShea (approved):

PR #500, opened 2023-03-22
JLBH for contentEquals. Allow perf tests to run on all Java versions. Closes #472

### LINKED ISSUE #472: Improve Bytes performance tests
1. ensure they show up in [TC statistics panel](https://teamcity.chronicle.software/project/Chronicle_ChronicleBytes?projectTab=stats&branch=%3Cdefault%3E#all-projects)
2. make sure all JLBH tests ran
3. extra specialisations of ContentEqualsJLBHTest for smaller size and also compare vs non-vectorized

JerryShea:
Also improves diagnosis around vectorizedMismatch and allows to be disabled with `disable.vectorized.content_equals`

### DIFF
--- a/src/test/java/net/openhft/chronicle/bytes/perf/ContentEqualJLBH.java
+++ b/src/test/java/net/openhft/chronicle/bytes/perf/ContentEqualJLBH.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright (c) 2016-2022 chronicle.software
+ *
+ *     https://chronicle.software
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package net.openhft.chronicle.bytes.perf;
+
+import net.openhft.chronicle.bytes.Bytes;
+import net.openhft.chronicle.bytes.BytesUtil;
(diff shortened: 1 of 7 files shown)


src/test/java/net/openhft/chronicle/bytes/perf/ContentEqualJLBH.java
```diff
@@ -0,0 +1,85 @@
...
+        String javaVersion = Jvm.isJava9Plus() ? Jvm.isJava15Plus() ? "17" : "11" : "8";
+        TeamCityHelper.teamCityStatsLastRun(getClass().getSimpleName() + "-Java" + javaVersion + "-" + bytes1.length(), jlbh, ITERATIONS, System.out);
+    }
+
+    @Override
+    public void run(long startTimeNS) {
+        boolean matching = BytesInternal.contentEqual(bytes1, bytes2);
+        if (matching)
+            throw new IllegalStateException();
+        jlbh.sampleNanos(System.nanoTime() - startTimeNS);
```
nicktindall, line 77: Would it be better to have a few calls to contentEquals that varied in equality (e.g. below) because sometimes the performance of the comparisons depends on how much of the string is equal?

|lhs              | rhs         |
|---------------|--------------|
|`xxxxxxx`   | `xxxxxxx`
|`xxxxxxx`   | `xxxxxxo`
|`xxxxxxx`   | `xxxxxoo`
|`xxxxxxx`   | `xxxxooo`
|`xxxxxxx`   | `xxxoooo`
|`xxxxxxx`   | `xxooooo`
|`xxxxxxx`   | `xoooooo`
|`xxxxxxx`   | `ooooooo`

nicktindall (commented):
Looks good but consider content being compared for equality, even a poor equals implementation can be fast if the first character in the string differs.

nicktindall (approved):

src/test/java/net/openhft/chronicle/bytes/perf/ContentEqualJLBH.java
JerryShea: Great idea - done

PR #501, opened 2023-03-22
Make Issue128Test not fail when the numbers differ due to one being written with scientific notation

nicktindall:
This is breaking the build currently

### DIFF
--- a/src/test/java/net/openhft/chronicle/bytes/Issue128Test.java
+++ b/src/test/java/net/openhft/chronicle/bytes/Issue128Test.java
@@ -34,10 +34,6 @@ public class Issue128Test extends BytesTestCommon {
         DF = df;
     }
 
-    static String toDecimal(double d) {
-        return DF.format(d);
-    }
-
     @Test
     public void testCorrect() {
         Bytes<?> bytes = Bytes.allocateDirect(32);
@@ -51,7 +47,7 @@ public class Issue128Test extends BytesTestCommon {
                 doTest(bytes, v7);
                 double v8 = (double) i / 100_000_000;
                 doTest(bytes, v8);
-                double v9 = (double) i / 100_000_000;
+                double v9 = (double) i / 1_000_000_000;
                 doTest(bytes, v9);
             }
         } finally {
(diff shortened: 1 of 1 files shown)


src/test/java/net/openhft/chronicle/bytes/Issue128Test.java
```diff
@@ -62,8 +58,13 @@ public void testCorrect() {
...
-            assertEquals(DF.format(v), output);
-//            System.out.println(DF.format(v)+" != " + output);
+        if (Double.parseDouble(output) != v || format.length() != output.length()) {
+            // Don't compare strings if we've added an exponent
+            if (!output.contains("E")) {
+                assertEquals(DF.format(v), output);
+            } else {
+                assertEquals(v, Double.parseDouble(output), 0.0);
+            }
+        }
```
nicktindall, line 68: We use `StringBuffer.append(..)` to render doubles below some threshold (currently `6E-8`) and that writes small numbers as scientific notation. That threshold may change in the future, so I don't want to bake it into this test, but rather just continue testing all values that don't end up being rendered as scientific notation.

src/test/java/net/openhft/chronicle/bytes/Issue128Test.java
```diff
@@ -51,7 +47,7 @@ public void testCorrect() {
                 doTest(bytes, v7);
                 double v8 = (double) i / 100_000_000;
                 doTest(bytes, v8);
-                double v9 = (double) i / 100_000_000;
+                double v9 = (double) i / 1_000_000_000;
```
nicktindall, line 50: This looks like it was a typo

src/test/java/net/openhft/chronicle/bytes/Issue128Test.java
```diff
@@ -34,10 +34,6 @@ public class Issue128Test extends BytesTestCommon {
         DF = df;
     }
 
-    static String toDecimal(double d) {
-        return DF.format(d);
-    }
```
nicktindall, line 39: unused code

JerryShea (approved):

PR #504, opened 2023-03-28
Feature/misc perf tweaks2

JerryShea:


src/main/java/net/openhft/chronicle/bytes/ByteStringAppender.java
```diff
@@ -224,35 +223,8 @@ default B append(double d)
     @NotNull
     default B append(double d, int decimalPlaces)
             throws BufferOverflowException, IllegalArgumentException, IllegalStateException, ArithmeticException {
-        if (decimalPlaces < 0)
```
JerryShea, line 227: moved to BytesInternal

peter-lawrey (approved):

PR #511, opened 2023-04-04
Test showing bugs with slices and MappedBytes

TwoClocks:
Add a unit test that shows bugs with MappedBytes, slices and moving them independently.

See  #512 for issue.


peter-lawrey:
The real bug here is that this method shouldn't conditionally produce a slice() esp when bytesForWrite() doesn't.
This requires some thought.

PR #516, opened 2023-05-24
added predicate to read one builder

RobAustin:


src/main/java/net/openhft/chronicle/bytes/MethodReaderBuilder.java
```diff
@@ -45,15 +48,21 @@ default MethodReaderBuilder exceptionHandlerOnUnknownMethod(ExceptionHandler exc
...
     MethodReaderBuilder metaDataHandler(Object... components);
 
     /**
      * Build a MethodReader using the following components to call
+     *
      * @param components to call
      * @return this
      */
     MethodReader build(Object... components);
+
```
JerryShea, line 64: javadoc please!

george-ball (approved):

PR #518, opened 2023-06-13
ensure that CommonMappedBytes.backingFileIsReadOnly is true if you open a readOnly view of a writeable file

JerryShea:
build-all-branch looks good

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/internal/CommonMappedBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/internal/CommonMappedBytes.java
@@ -73,7 +73,7 @@ public abstract class CommonMappedBytes extends MappedBytes {
         assert mappedFile != null;
         this.mappedFile = mappedFile;
         mappedFile.reserve(this);
-        this.backingFileIsReadOnly = !mappedFile.file().canWrite();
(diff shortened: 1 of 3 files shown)


src/main/java/net/openhft/chronicle/bytes/internal/CommonMappedBytes.java
```diff
@@ -73,7 +73,7 @@ protected CommonMappedBytes(@NotNull final MappedFile mappedFile, final String n
         assert mappedFile != null;
         this.mappedFile = mappedFile;
         mappedFile.reserve(this);
-        this.backingFileIsReadOnly = !mappedFile.file().canWrite();
+        this.backingFileIsReadOnly = mappedFile.readOnly();
```
nicktindall, line 76: So this will check if the mapped file was mapped in read-only mode, should we also check that the behaviour is as expected for files that are marked read-only by the file-system?

Or at least confirm that you can't create a MappedFile in read-write mode when it's read-only on disk?

JerryShea: OK, done

nicktindall (approved):

PR #520, opened 2023-06-20
Reduce any race condition on closing the underlying bytes

peter-lawrey:


src/test/java/net/openhft/chronicle/bytes/NotNullInstrumentationDisabledTest.java
```diff
@@ -1,39 +0,0 @@
-/*
- * Copyright (c) 2016-2022 chronicle.software
```
JerryShea, line 2: why delete this?

peter-lawrey: There was another test that failed if annotation processing was turned on and this test highlighted the problem.
With this PR, that test runs fine with or without annotation processing.

JerryShea (approved):

Issue #523, opened 2023-06-22
Bytes.append(double) output scientific notation (more eager than expected)


alamar:
The behaviour of `AbstractBytes.append(double)` has changed in #502 (PR #507)

Previously, `bytes.append(2e-8)` would append `0.00000002`. It now uses scientific notation.

The method is documented in ByteStringAppender as “Appends a double in decimal notation”.

The difficulties are the following:
For some double values, it would become untenable to render them in decimal notation, as they may become circa 400 decimal positions long, so we will likely still drop to scientific notation at some point (and update the javadoc, hopefully).
We also want to avoid showing quantization artifacts and continuous fractions for a subset of human readable fractions, e.g., we do not want to render `0.0009` as `"0.0008999999999999998"`.
However, the actual limits used are also subject to debate. A suggestion is rounding to between 15 and 18 significant figures if possible instead of switching to scientific notation.

alamar:
Current behavior is due to the use of Core's `net.openhft.chronicle.core.io.UnsafeText#appendDouble`

JerryShea:
Fixed by #526 #529 #530

PR #526, opened 2023-06-23
Bytes/issues/523

peter-lawrey:
fixes https://github.com/OpenHFT/Chronicle-Bytes/issues/523 see also https://github.com/ChronicleEnterprise/Chronicle-FIX/pull/1785 which test this in Chronicle FIX


pom.xml
```diff
@@ -29,7 +29,7 @@
     </parent>
 
     <artifactId>chronicle-bytes</artifactId>
-    <version>2.24ea13-SNAPSHOT</version>
+    <version>2.24ea14-SNAPSHOT</version>
```
JerryShea: ?

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
```diff
@@ -1329,6 +1329,240 @@ public boolean lenient() {
         return base.lenient();
     }
 
+    @Override
+    public @NotNull Bytes<Void> append(char ch) throws IllegalStateException {
```
JerryShea, line 1335: why do we need new methods in HexDumpBytes?

src/main/java/net/openhft/chronicle/bytes/MappedBytes.java
```diff
@@ -36,7 +36,7 @@
  * <p>
  * Memory is grouped in chunks of 64 MB by default.
  * The chunk size can be increased greatly if the OS.isSparseFileSupported()
- * e.g. blockSize(512 << 30)
+ * e.g. blockSize(512 &gt;&gt; 30)
```
JerryShea: `&lt;`

src/main/java/net/openhft/chronicle/bytes/UncheckedNativeBytes.java
```diff
@@ -908,6 +914,59 @@ public Bytes<U> appendUtf8(char[] chars, @NonNegative int offset, @NonNegative i
...
+        long pos2 = writePosition()-1;
+        while (pos2 > pos) {
+            byte b1 = bytesStore.readByte(pos);
+            byte b2 = bytesStore.readByte(pos2);
+            bytesStore.writeByte(pos, b2);
+            bytesStore.writeByte(pos2, b1);
+            pos++;
+            pos2--;
+        }
+    }
```
nicktindall: Could this also live in decimalizer? it seems to be repeated in AbstractBytes and UncheckedNativeBytes

src/main/java/net/openhft/chronicle/bytes/internal/Decimalizer.java
```diff
@@ -0,0 +1,217 @@
...
+
+        /**
+         * Converts a float value to its decimal representation using a simple rounding approach and appends it.
+         *
+         * @param value           the float value to be converted.
+         * @param decimalAppender the DecimalAppender used to store the converted decimal value.
+         * @return true if the conversion and appending was successful, false otherwise.
+         */
+        public boolean toDecimal(float value, DecimalAppender decimalAppender) {
+            boolean sign = Double.doubleToLongBits(value) < 0;
```
nicktindall: Use `Float.floatToIntBits(...)` to avoid the cast perhaps?

src/main/java/net/openhft/chronicle/bytes/internal/Decimalizer.java
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * BigDecimal-based implementation of Decimalizer for higher precision conversions.
+     */
+    class UsesBigDecimal implements Decimalizer {
+
+        /**
+         * Field reference to the internal storage of BigDecimal values.
+         */
+        static final java.lang.reflect.Field INT_COMPACT = Jvm.getFieldOrNull(BigDecimal.class, "intCompact");
```
nicktindall: Not worth using a method handle for this access?

src/test/java/net/openhft/chronicle/bytes/BytesTest.java
```diff
@@ -1071,21 +1072,158 @@ public void isClear() {
         bytes.releaseLast();
     }
 
+    @Test
+    public void testAppendDouble() {
+
+        // TODO FIX
```
JerryShea: please fix!

src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
peter-lawrey: HexDumpBytes records the bytes written and the structure of how they were written.
This was done for binary writing but not text writing.

pom.xml
peter-lawrey: The current version is 2.24ea14-SNAPSHOT as reflected in the chronicle-bom.
If you build version 2.24ea13-SNAPSHOT locally, you won't be using that version because the bom says to use 2.24ea14-SNAPSHOT

src/main/java/net/openhft/chronicle/bytes/MappedBytes.java
peter-lawrey: Using < in Javadoc isn't valid.

src/main/java/net/openhft/chronicle/bytes/UncheckedNativeBytes.java
peter-lawrey: It repeats the same structure but using it in Decimaliser would turn every byte into a polymorphic call.
There is a benchmark so we could explore how much difference it makes.

src/main/java/net/openhft/chronicle/bytes/internal/Decimalizer.java
peter-lawrey: Indeed.

src/main/java/net/openhft/chronicle/bytes/internal/Decimalizer.java
peter-lawrey: I could explore if that makes a difference. Using Unsafe is probably the fastest, however, given this is using BigDecimal at all isn't the most performant.

pom.xml
nicktindall: It's actually weird that this is showing up in the diff, because according to my local copy it is already `2.24ea14-SNAPSHOT` on develop

JerryShea: Maybe just need to rebase?

src/main/java/net/openhft/chronicle/bytes/MappedBytes.java
JerryShea: No but you have changed < to >

src/test/java/net/openhft/chronicle/bytes/BytesTest.java
peter-lawrey: This due to parseDouble() so needs another issue

src/main/java/net/openhft/chronicle/bytes/BytesMethodReaderBuilder.java
```diff
@@ -41,6 +41,13 @@ public MethodReaderBuilder exceptionHandlerOnUnknownMethod(ExceptionHandler exce
         return this;
     }
 
+    @Deprecated(/* remove in x.25 */)
+    @Override
+    public MethodReaderBuilder warnMissing(boolean warnMissing) {
```
peter-lawrey, line 46: removed prematurely

src/test/java/net/openhft/chronicle/bytes/BytesTest.java
peter-lawrey: https://github.com/OpenHFT/Chronicle-Bytes/issues/527

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -234,25 +236,77 @@ public boolean compareAndSwapLong(@NonNegative long offset, long expected, long
...
     @Override
     public @NotNull AbstractBytes<U> append(double d)
             throws BufferOverflowException, IllegalStateException {
-        boolean fits = canWriteDirect(32);
-        if (fits) {
-            long address = addressForWrite(writePosition());
-            long address2 = UnsafeText.appendDouble(address, d);
-            writeSkip(address2 - address);
+        if (X_23_APPEND_DOUBLE) {
+            boolean fits = canWriteDirect(32);
```
JerryShea: can we extract the contents of the if to a function as it likely wont be called

src/main/java/net/openhft/chronicle/bytes/ByteStringAppender.java
```diff
@@ -395,5 +395,6 @@ default B append(@NotNull BigDecimal bigDecimal) {
     }
 
     // internal method to cache a byte[]
+    @Deprecated(/* to be removed in x.26 */)
```
JerryShea: can we remove sooner? I vote for x.25 or even x.24

src/test/java/net/openhft/chronicle/bytes/Issue523Test.java
```diff
@@ -0,0 +1,86 @@
...
+                            for (int j = 0; j < 1000; j++) {
+                                int k = i + j * 1000;
+                                double d = k / ee;
+                                bytes.clear();
+                                bytes.append(d);
+                                int l = (int) bytes.readRemaining();
+                                if (l > 2 + e)
+                                    return d + ": " + bytes + " too long";
+/*
+                                for (int ii = 0; ii < l; ii++)
```
JerryShea: pls remove commented-out code

src/main/java/net/openhft/chronicle/bytes/ByteStringAppender.java
peter-lawrey: This is used by a number of methods

peter-lawrey:
Discussed these changes and inclusion in x.23 with Jerry this morning

PR #536, opened 2023-07-10
Added javadoc 202307

peter-lawrey:


JanStureNielsen:
This looks like multiple independent changes including:

  1. JavaDoc additions
  2. Remove try-catch of ISE
  3. BUE, BOE, AIE changes
  4. Logic changes/cleanup

peter-lawrey:
This code needed a git rebase with a change that was still a PR.

There is an apparent ligc change in Bytes.indexOf however, the indenting has changed and this confused the comparator.

Some RuntimeExceptions were dropped from the throws clause as it was apparent from doing the Javadoc that these never actually occur (or if they did, it wasn't know what they would mean)

PR #537, opened 2023-07-13
update javadoc for RB stats

JerryShea:


src/main/java/net/openhft/chronicle/bytes/MappedBytes.java
```diff
@@ -34,7 +34,7 @@
  * A specialized implementation of {@link AbstractBytes} that wraps memory-mapped data for efficient random file access.
  *
  * <p>Memory is grouped in chunks of 64 MB by default. The chunk size can be significantly increased if the
- * OS supports sparse files via the OS.isSparseFileSupported() method, e.g. blockSize(512 << 30).
+ * OS supports sparse files via the OS.isSparseFileSupported() method, e.g. {@code blockSize(512 << 30)}.
```
nicktindall: could `{@link OS#isSparseFileSupported()}` here

nicktindall (approved):

src/main/java/net/openhft/chronicle/bytes/MappedBytes.java
JerryShea: thanks. done

PR #549, opened 2023-08-14
Tests/guarded native bytes

peter-lawrey:


peter-lawrey:
Replaced by a newer PR

PR #551, opened 2023-08-15
Fix an issue where Bytes.compact() doesn't do anything if readRemaini…

peter-lawrey:
…ng() == 0

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
@@ -128,15 +128,19 @@ public abstract class AbstractBytes<U>
     public Bytes<U> compact()
             throws ClosedIllegalStateException, ThreadingIllegalStateException {
         assert DISABLE_SINGLE_THREADED_CHECK || threadSafetyCheck(true);
-        long start = start();
-        long readRemaining = readRemaining();
-        if ((readRemaining > 0) && (start < readPosition)) {
-            bytesStore.move(readPosition, start, readRemaining);
-        }
-        readPosition = start;
-        uncheckedWritePosition(start + Math.max(0, readRemaining));
+        try {
+            long start = start();
+            long readRemaining = readRemaining();
+            if ((readRemaining > 0) && (start < readPosition)) {
+                bytesStore.move(readPosition, start, readRemaining);
+            }
+            readPosition = start;
+            uncheckedWritePosition(start + Math.max(0, readRemaining));
 
-        return this;
+            return this;
+        } catch (IllegalStateException ignored) {
+            return this;
+        }
     }
 
     @Override
--- a/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
+++ b/src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
@@ -127,16 +127,33 @@ public abstract class AbstractBytes<U>
     @Override
     public Bytes<U> compact()
             throws ClosedIllegalStateException, ThreadingIllegalStateException {
+
+        // Asserts that the current method is thread-safe or that single-threaded checks are disabled
         assert DISABLE_SINGLE_THREADED_CHECK || threadSafetyCheck(true);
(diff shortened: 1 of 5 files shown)


src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -130,11 +130,11 @@ public Bytes<U> compact()
         assert DISABLE_SINGLE_THREADED_CHECK || threadSafetyCheck(true);
         long start = start();
         long readRemaining = readRemaining();
-        if ((readRemaining > 0) && (start < readPosition)) {
+        if ((readRemaining > 0) && (start < readPosition))
```
JerryShea: can we leave the curly brackets?

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -125,17 +125,21 @@ public void move(@NonNegative long from, @NonNegative long to, @NonNegative long
...
+            long readRemaining = readRemaining();
+            if ((readRemaining > 0) && (start < readPosition)) {
+                bytesStore.move(readPosition, start, readRemaining);
+            }
             readPosition = start;
-            uncheckedWritePosition(readPosition + readRemaining);
+            uncheckedWritePosition(start + Math.max(0, readRemaining));
+
+            return this;
+        } catch (IllegalStateException ignored) {
```
JerryShea: should be ClosedIllegalStateException? Should we have a test for calling compact on a released Bytes?

src/main/java/net/openhft/chronicle/bytes/Bytes.java
```diff
@@ -998,14 +998,13 @@ default boolean isEqual(@Nullable String other)
      * <p>
      * This operation is useful to free up space for writing by discarding bytes that have already
      * been read.
-     * 
+     * <p>
+     * If the underlying Bytes have been released, this does nothing.
      *
      * @return This Bytes object, for method chaining.
-     * @throws IllegalStateException if this Bytes object has been previously released.
```
JerryShea: why change this?

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
peter-lawrey: compact() should fail fast.
While it could ignore being called after closing, this is almost certainly a coding error.

src/main/java/net/openhft/chronicle/bytes/Bytes.java
peter-lawrey: This was changed to reflect the actual implementation.
However, the implementation should fail fast and this has been covered by another change.

src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
JerryShea: Don't understand.... If we want to fail fast, why are we catching and ignoring the ISE?

src/main/java/net/openhft/chronicle/bytes/Bytes.java
```diff
@@ -990,11 +990,13 @@ default boolean isEqual(@Nullable String other)
...
     /**
-     * Compacts this Bytes object by moving the read position to the start.
+     * Compacts this Bytes object by moving the readPosition() to the start(),
+     * adjusting the readLimit()/writePosition() appropriately so that the readRemaining() is unchanged.
      * <p>
      * This operation is useful to free up space for writing by discarding bytes that have already
      * been read.
-     *
+     * <p>
+     * NOTE: If the space freed in longs is less than 1/4 the readRemaining in longs, this method does nothing to reduce movement overhead
```
JerryShea, line 999: "longs" here is confusing

"If the space freed would be less than 1/4 of the amount to be moved, this method does nothing"

JerryShea (approved):
Approved but please take on board my comment re: javadoc and my DM comments

PR #554, opened 2023-08-30
Java 21 support

peter-lawrey:


RobAustin (approved):

RobAustin:
Note the bug reported in SonarCloud
image
 its says

When arithmetic is performed on integers, the result will always be an integer. You can assign that result to a long, double, or float with automatic type conversion, but having started as an int or long, the result will likely not be what you expect.
For instance, if the result of int division is assigned to a floating-point variable, precision will have been lost before the assignment. Likewise, if the result of multiplication is assigned to a long, it may have already overflowed before the assignment.
In either case, the result will not be what was expected. Instead, at least one operand should be cast or promoted to the final type before the operation takes place.

peter-lawrey:
It's right, except a byte[] can only have an int number of bytes.

PR #555, opened 2023-08-30
Improve support for GuardedNativeBytes

peter-lawrey:


src/test/java/net/openhft/chronicle/bytes/GuardedNativeBytesTest.java
```diff
@@ -0,0 +1,99 @@
+/*
```
RobAustin, line 1: I suggest with javadoc below 

```

import org.junit.Test;
import static org.junit.Assert.assertEquals;
import static org.junit.Assert.assertTrue;

/**
 * This class contains JUnit test methods for testing the behavior
 * of the GuardedNativeBytes class. 
 * <p>
 * It aims to test various primitive data types and their conversions
 * using the GuardedNativeBytes class.
 * </p>
 */
public class GuardedNativeBytesTest {

    /**
     * Tests the reading and writing of various binary primitives.
     * <p>
     * This test method performs the following steps:
     * <ul>
     *   <li>Writes different types of binary data into a GuardedNativeBytes object.</li>
     *   <li>Checks the generated hexadecimal string against an expected value.</li>
     *   <li>Reads the binary data back and checks that it matches the original input.</li>
     * </ul>
     * </p>
     */
    @Test
    public void testBinaryPrimitive() {
        final GuardedNativeBytes<?> bytes = new GuardedNativeBytes(new HexDumpBytes(), 256);
        try {
            // Writing various binary primitives to the GuardedNativeBytes object.
            bytes.writeHexDumpDescription("flag").writeBoolean(true);
            bytes.writeHexDumpDescription("s8").writeByte((byte) 1);
            bytes.writeHexDumpDescription("u8").writeUnsignedByte(2);
            // ... (more write operations here)
            // Check the generated hex string.
            final String expected = /* ... */;
            final String actual = bytes.toHexString();
            assertEquals(expected, actual);

            // Reading the data back and checking it.
            final boolean flag = bytes.readBoolean();
            // ... (more read operations and assertions here)
        } finally {
            bytes.releaseLast();
        }
    }
}
```

RobAustin (approved):

PR #559, opened 2023-09-19
Scoped thread locals

nicktindall:
Use scoped thread locals in Bytes

Depends on https://github.com/OpenHFT/Chronicle-Core/pull/576


pom.xml
```diff
@@ -29,7 +29,7 @@
     </parent>
 
     <artifactId>chronicle-bytes</artifactId>
-    <version>2.24ea17-SNAPSHOT</version>
+    <version>2.24ea18-SNAPSHOT</version>
```
nicktindall: 2.24ea17 seems to have gone missing

JerryShea: I think we made a release yesterday so maybe need to rebase

JerryShea (approved):

PR #562, opened 2023-10-10
Extend API to support custom page size (hugetlbfs).

yevgenp:
Part of https://github.com/ChronicleSoftware/QueueEvolution/issues/68

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/PageUtil.java
+++ b/src/main/java/net/openhft/chronicle/bytes/PageUtil.java
@@ -58,7 +58,7 @@ public final class PageUtil {
         }
     }
 
-    private static List<String> readMountInfo(String path) {
+    static List<String> readMountInfo(String path) {
         try {
             return Files.readAllLines(Paths.get(path));
         } catch (IOException e) {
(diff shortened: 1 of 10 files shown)


src/main/java/net/openhft/chronicle/bytes/PageUtil.java
```diff
@@ -0,0 +1,146 @@
...
+    private static void insert(String path, int size) {
+        TrieNode curr = root;
+        for (String dir : path.split("/")) {
+            if (dir.isEmpty()) continue;
+            curr.childs.putIfAbsent(dir, new TrieNode());
+            curr = curr.childs.get(dir);
+        }
+        curr.isLeaf = true;
+        curr.pageSize = size;
+        Jvm.warn().on(PageUtil.class, format("Found pageSize={0} for mount point ''{1}''", size, path));
```
tgd: Please could we change this to info level. Its presence in some Queue tests causes them to fail as the tests are configured to fail on warnings (we could ignore them if preferred).

I don't believe there is a Jvm.info() though and it might be outside of the scope of this change to add it. Let me know what you think

yevgenp: Changed to info (apparently it's called PERF).

peter-lawrey (approved):

PR #566, opened 2023-10-20
Enable page size autoconfiguration.

yevgenp:


tgd (approved):

nicktindall (approved):
This is a vastly superior approach IMO

tgd:
FYI: Last week the path to hugetlbfs changed from `/huge` to `/mnt/huge` inside the docker container. Any tests that used to reference `/huge` explicitly need to be changed to use `/mnt/huge` instead.

PR #568, opened 2023-10-24
Deprecate code due x.25.

yevgenp:


src/main/java/net/openhft/chronicle/bytes/AbstractBytes.java
```diff
@@ -1527,7 +1525,6 @@ public int byteCheckSum(@NonNegative int start, @NonNegative int end)
         return sum & 0xFF;
     }
 
-    @Override
     public boolean isImmutableEmptyByteStore() {
         return bytesStore.capacity() == 0;
     }
```
nicktindall, line 1530: Should this method be removed altogether? @peter-lawrey

yevgenp: Let's not hold the deprecation that has to be done for all products by tomorrow because of such little issue.

nicktindall: I think we should, otherwise it'll hang around forever. When the removal date arrives after a deprecation we remove.

peter-lawrey (approved):

PR #571, opened 2023-10-30
Fix deprecated classes for x.25

peter-lawrey:


yevgenp (approved):

src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java
```diff
@@ -138,7 +138,7 @@ enum BytesInternal {
             }
         } catch (Exception e) {
             if (e.getClass().getName().equals("java.lang.reflect.InaccessibleObjectException"))
-                Jvm.warn().on(BytesInternal.class, "Cannot get access to vectorizedMismatch. The following command line args are required: " +
+                Jvm.debug().on(BytesInternal.class, "Cannot get access to vectorizedMismatch. The following command line args are required: " +
```
JerryShea, line 141: I don't agree with this - this is a genuine warning

PR #580, opened 2023-11-24
Deprecated appendBytesFromStart

chroniclekevinpowe:
Deprecated method as part of removing use from Chronicle FIX.

Appending the unicode character \u2016 to the exception message causes issues when the exception message is used as part of generating a Reject message. The Reject message is treated as a single-byte character string when being output, create a garbled reject message.

### DIFF
--- a/src/main/java/net/openhft/chronicle/bytes/BytesUtil.java
+++ b/src/main/java/net/openhft/chronicle/bytes/BytesUtil.java
@@ -518,7 +518,7 @@ public enum BytesUtil {
      * @throws ClosedIllegalStateException    If the resource has been released or closed.
      * @throws ThreadingIllegalStateException If this resource was accessed by multiple threads in an unsafe way.
      */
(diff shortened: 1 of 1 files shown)


JerryShea (approved):

src/main/java/net/openhft/chronicle/bytes/BytesUtil.java
```diff
@@ -517,6 +517,7 @@ public static void appendUtf8(@NotNull StreamingDataOutput out, @NotNull CharSeq
      * @throws ThreadingIllegalStateException If this resource was accessed by multiple threads in an unsafe way.
      */
     // used by Chronicle FIX.
```
JerryShea: pls remove this line!

chroniclekevinpowe: Done!

nicktindall (approved):

PR #581, opened 2023-11-24
Fix issues after method removal, discovered alongside

alamar:
see OpenHFT/Chronicle-Core#602


alamar:
https://teamcity.chronicle.software/buildConfiguration/Chronicle_BuildAll_BumpedBranch/882236?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true is green

PR #585, opened 2024-01-04
Deprecate StringUtils.setCount as not compatible with Java 21.

yevgenp:


peter-lawrey (approved):

yevgenp:
`StringUtils.setCount` can be left as it is inside **Bytes** as it's complemented with extra logic in `BytesInternal`.
Which was not the case for Wires.

PR #588, opened 2024-01-15
Added test for ISO-8859-1 encoding ensuring all printable characters …

JerryShea:
…converted correctly.

Added Javadoc for Bytes.toString indicating character support Exposed to8BitString


nicktindall (approved):

JerryShea:
Actually I won't backport this, it is not strictly necessary

PR #591, opened 2024-01-15
remove static to8BitString as already implemented as member method

JerryShea:


src/main/java/net/openhft/chronicle/bytes/Bytes.java
```diff
@@ -644,18 +644,6 @@ static String toString(@NotNull final Bytes<?> buffer,
...
-     * <p>Extracts a string from the provided {@code bytes}. The store's state
-     * remains unchanged by this method.</p>
-     *
-     * <p>This method supports all characters in the ISO-8859-1 encoding scheme.</p>
-     *
-     * @param bytes BytesStore to extract string from
-     * @return A string extracted from the BytesStore
-     * @throws ClosedIllegalStateException if the BytesStore provided has been previously released
-     */
-    static String to8bitString(@NotNull BytesStore bytes) throws ClosedIllegalStateException { return BytesInternal.to8bitString(bytes); }
```
JerryShea, line 657: not required to be added. Bytes(Store) has a member function already

chroniclekevinpowe (approved):
Thank you for rolling this back!

PR #599, opened 2024-02-08
Ticket/1808

peter-lawrey:


yevgenp (approved):
Could you add a test case from the customer's ticket i.e. with 10 digits precision just to be sure?

tgd (approved):

tgd:
> Could you add a test case from the customer's ticket i.e. with 10 digits precision just to be sure?

+1

Peter, the following test can be added, and will pass now:

```java
@Test
void acceptanceTest() {
  double value = 584600.0;
  Bytes<?> bytes = Bytes.elasticHeapByteBuffer().append(value, 10);
  assertEquals(String.format("%.10f", value), bytes.toString());
}
```
Peter, I raised ticket https://github.com/OpenHFT/Chronicle-Bytes/issues/600 please can you reference it when you commit.

Issue #604, opened 2024-03-07
Bytes contentEquals of heap backed byte instances and readSkip


tgd:
# Steps to reproduce

1. Create two bytes instance - one direct and one heap based
2. Advance the read position on both, the same amount, by calling `readSkip`
3. Compare the equality of the bytes instances via `contentEquals`
4. The two buffers, containing the same data will not be considered equal

# Root cause

`BytesInternal#contentEquals` eventually ends up calling `java11ContentEqualUsingVectorizedMismatch` in certain situations. In situations where this gets called and readSkip is called on a heap bytes instance the read offset of the heap bytes instance will not actually be advanced before the comparison takes place. This results in two bytes instances being compared with their read offsets set to different positions.

# Test cases to reproduce

More in the linked PR.

```java
        @ParameterizedTest
        @MethodSource("bufferArguments")
        void longEqualsWithReadSkip(Bytes<?> left, Bytes<?> right) {
            left.write("abcdefghijzlmnopqrstuvwxyz".getBytes());
            left.readSkip(8);
            right.write("abcdefghijzlmnopqrstuvwxyz".getBytes());
            right.readSkip(8);
            assertEquals(left, right);
        }

        public Stream<Arguments> bufferArguments() {
            return Stream.of(
                    Arguments.of(Bytes.allocateElasticOnHeap(), Bytes.allocateElasticDirect()), // heap, direct
                    Arguments.of(Bytes.allocateElasticDirect(), Bytes.allocateElasticOnHeap()), // direct, heap
                    Arguments.of(Bytes.allocateElasticDirect(), Bytes.allocateElasticDirect()), // direct, direct
                    Arguments.of(Bytes.allocateElasticOnHeap(), Bytes.allocateElasticOnHeap()) // heap, heap
            );
        }
```

hft-team-city:
Released in [Chronicle-Bytes-2.23.44](https://github.com/OpenHFT/Chronicle-Bytes/releases/chronicle-bytes-2.23.44), [BOM-2.23.216](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.23.216)

hft-team-city:
Released in [Chronicle-Bytes-2.25ea8](https://github.com/OpenHFT/Chronicle-Bytes/releases/chronicle-bytes-2.25ea8), [BOM-2.25ea32](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.25ea32)

PR #605, opened 2024-03-07
 Bytes contentEquals of heap backed byte instances and readSkip

tgd:
Please see here for write up - https://github.com/OpenHFT/Chronicle-Bytes/issues/604


JerryShea (approved):
great change thanks Tom

Issue #610, opened 2024-03-14
MappedBytesStore.sync() silently fails due to a bug


peter-lawrey:

tgd:
Hi @peter-lawrey what's the bug?

tgd:
Peter I believed this is fixed by the linked PRs. Please re-open if you disagree.

PR #611, opened 2024-03-14
MappedBytesStore.sync() silently fails due to a bug. #610

peter-lawrey:


tgd:
Hi @peter-lawrey what's the bug?

peter-lawrey:
The bug is that the starting address has to be a multiple of a page; if it isn't, the method returns errno -1, Invalid Argument.

tgd (approved):

JerryShea (approved):
change looks good but I echo @tgd 's comment - what's the bug? Can we change the description?

PR #614, opened 2024-03-18
Bytes/issues/612

peter-lawrey:


tgd (approved):

src/main/java/net/openhft/chronicle/bytes/MappedBytesStore.java
```diff
@@ -479,19 +486,26 @@ public void syncMode(SyncMode syncMode) {
...
+     *
+     * @param position to sync with the syncMode()
+     * @param syncMode to use
+     */
+    public void syncUpTo(long position, SyncMode syncMode) {
+        if (syncMode == SyncMode.NONE || address == 0 || refCount() <= 0 || !OS.isLinux())
             return;
-        long length = position - start;
-        if (length <= syncLength)
+        long positionFromStart = Math.min(safeLimit, position) - start;
```
tgd: Hi @peter-lawrey if you have time it would be good to get a new test around this (unless its covered already)

peter-lawrey: Good catch, more tests needed.

PR #615, opened 2024-03-19
Fix issue with read offset being ignored

alamar:
Add tests
see OpenHFT/Chronicle-Core#602

### DIFF
--- a/src/test/java/net/openhft/chronicle/bytes/HeapByteStoreTest.java
+++ b/src/test/java/net/openhft/chronicle/bytes/HeapByteStoreTest.java
@@ -21,6 +21,11 @@ import net.openhft.chronicle.bytes.internal.HeapBytesStore;
 import org.jetbrains.annotations.NotNull;
 import org.junit.Test;
 
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.charset.StandardCharsets;
+import java.util.concurrent.Callable;
+
 import static org.junit.Assert.*;
 
 public class HeapByteStoreTest extends BytesTestCommon {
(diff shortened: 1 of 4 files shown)


alamar:
@peter-lawrey @JerryShea WDYT?

src/test/java/net/openhft/chronicle/bytes/readme/PrimitiveTest.java
```diff
@@ -221,9 +221,22 @@ public void testBinaryPrimitiveOffset() {
     }
 
     @Test
-    public void testTextPrimitive() {
+    public void testTextPrimitiveByteBuffer() {
+        doTestTextPrimitive(Bytes.elasticHeapByteBuffer(64));
```
JerryShea, line 225: these tests do not break if I revert the change HeapBytesStore

JerryShea (approved):

src/test/java/net/openhft/chronicle/bytes/HeapByteStoreTest.java
```diff
@@ -48,4 +53,35 @@ public void testElasticBytesEnsuringCapacity() {
...
+    public void testWriteOnHeap() throws Exception {
+        doTestWrite(() -> ByteBuffer.allocate(128));
+    }
+
+    @Test
+    public void testWriteDirect() throws Exception {
+        doTestWrite(() -> ByteBuffer.allocateDirect(128));
+    }
+
+    private void doTestWrite(Callable<ByteBuffer> generator) throws Exception {
```
JerryShea: consider moving this test to BytesTest so it tests this behaviour is checked for all Bytes types

src/test/java/net/openhft/chronicle/bytes/readme/PrimitiveTest.java
alamar: They do, but I still think more code coverage do not hurt. They used to be broken by something else which since have been taken care of, so I believe.

JerryShea: Fair enough!

JerryShea (approved):

JerryShea:
This is a great catch

PR #616, opened 2024-03-21
Delegate whether MSync occurs to the PosixAPI

peter-lawrey:


src/main/java/net/openhft/chronicle/bytes/MappedBytesStore.java
```diff
@@ -479,7 +479,7 @@ public void syncMode(SyncMode syncMode) {
      * @param position to sync with the syncMode()
      */
     public void syncUpTo(long position) {
-        if (syncMode == SyncMode.NONE || address == 0 || refCount() <= 0)
+        if (syncMode == SyncMode.NONE || address == 0 || refCount() <= 0 || !OS.isLinux())
```
tgd: Hi @peter-lawrey, you may be able to replace with `!isUnix` or alternatively `OS.isWindows()` (don't believe there is an `isUnix` in core). In other words this can likely work on Linux, Mac, other Unix like OS supporting `msync`.

Here's the code in the `Posix` API that switches the underlying Posix wrapper depending on platform:

https://github.com/OpenHFT/Posix/blob/ea/src/main/java/net/openhft/posix/internal/PosixAPIHolder.java#L17-L19

```java
posixAPI = Platform.getNativePlatform().isUnix()
        ? new JNRPosixAPI()
        : new WinJNRPosixAPI();
```

This ultimately resolves down to:

```java
    public final boolean isUnix() {
        return this.os != Platform.OS.WINDOWS;
    }
```

tgd (approved):

src/main/java/net/openhft/chronicle/bytes/MappedBytesStore.java
```diff
@@ -27,6 +27,7 @@
 import net.openhft.chronicle.core.io.ReferenceOwner;
 import net.openhft.chronicle.core.io.ThreadingIllegalStateException;
 import net.openhft.posix.PosixAPI;
+import net.openhft.posix.internal.jnr.WinJNRPosixAPI;
```
tgd, line 30: Are we calling anything from this?

PR #618, opened 2024-03-22
Error on deprecated

peter-lawrey:
This needs the PR https://github.com/OpenHFT/Chronicle-Core/pull/637 to pass

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -407,9 +407,9 @@
 
     <profiles>
         <profile>
-            <id>java11-release8</id>
+            <id>java9-release8</id>
             <activation>
-                <jdk>[11,</jdk>
+                <jdk>[9,)</jdk>
             </activation>
             <properties>
                 <maven.compiler.release>8</maven.compiler.release>
--- a/pom.xml
+++ b/pom.xml
@@ -18,14 +18,15 @@
     limitations under the License.
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
         <groupId>net.openhft</groupId>
         <artifactId>java-parent-pom</artifactId>
         <version>1.25.4</version>
-        <relativePath />
+        <relativePath/>
     </parent>
 
     <artifactId>chronicle-bytes</artifactId>
@@ -246,7 +247,7 @@
                     <source>1.8</source>
                     <target>1.8</target>
                     <encoding>UTF-8</encoding>
-                    <!--                    <verbose>true</verbose>-->
+                    <failOnWarning>true</failOnWarning>
                 </configuration>
             </plugin>
             <!--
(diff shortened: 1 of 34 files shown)


src/test/java/net/openhft/chronicle/bytes/internal/UnsafeTextTest.java
```diff
@@ -0,0 +1,235 @@
+/*
```
tgd, line 1: Added by accident?

src/main/java/net/openhft/chronicle/bytes/internal/UnsafeText.java
```diff
@@ -0,0 +1,308 @@
...
+import net.openhft.chronicle.core.*;
+
+import static net.openhft.chronicle.core.UnsafeMemory.MEMORY;
+import static net.openhft.chronicle.core.UnsafeMemory.UNSAFE;
+
+/**
+ * These are fast, unsafe ways to render text.
+ * NOTE: The caller has to ensure there is always plenty of memory to perform this operation.
+ */
+public final class UnsafeText {
```
tgd, line 29: Hi Peter is this a brand new class or refactored from something else?

tgd (approved):
Hi Peter - approved it but with a few questions whether some unrelated changes were added

src/test/java/net/openhft/chronicle/bytes/internal/UnsafeTextTest.java
peter-lawrey: The class this tests was moved from Core to Bytes.

src/main/java/net/openhft/chronicle/bytes/internal/UnsafeText.java
peter-lawrey: Moved from Core to Bytes to make it internal to Bytes, and eventually phase it out.

pom.xml
```diff
@@ -405,6 +406,15 @@
     </scm>
 
     <profiles>
+        <profile>
+            <id>java11-release8</id>
+            <activation>
+                <jdk>[11,</jdk>
```
nicktindall: This is not a valid range https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html

Should be `[11,)`

src/main/java/net/openhft/chronicle/bytes/BytesUtil.java
```diff
@@ -131,6 +131,7 @@ static int[] isTriviallyCopyable0(@NotNull Class<?> clazz) {
      * @param fields The fields of the class.
      * @return An array of two integers, where the first element is the minimum offset, and the second element is the maximum offset.
      */
+    @SuppressWarnings("deprecation")
```
nicktindall: We shouldn't be suppressing deprecation warnings, that will make these things very hard to find down the track. Unless the method itself is due to go at the same time as the deprecated thing it references?

src/main/java/net/openhft/chronicle/bytes/internal/ByteBuffers.java
```diff
@@ -0,0 +1,67 @@
...
+import java.nio.ByteBuffer;
+
+/**
+ * Utility class for working with {@link ByteBuffer} instances.
+ * <p>
+ * Provides functionality to directly set the address and capacity of a ByteBuffer.
+ * <p>
+ * Note: This class is deprecated and will be removed in version x.26.
+ */
+public final class ByteBuffers {
```
nicktindall, line 31: If this class is deprecated, can we add a deprecated annotation.

Also add `@deprecated` in the javadoc with a recommended alternative.

nicktindall (approved):
Approved, but we should make sure anything deprecated is marked/documented as such and be careful about suppressing deprecation warnings as it'll cause us headaches later on when we remove those things.

src/main/java/net/openhft/chronicle/bytes/internal/ByteBuffers.java
peter-lawrey: The original was deprecated, but not this one.

src/main/java/net/openhft/chronicle/bytes/BytesUtil.java
peter-lawrey: I have narrowed the SuppressWarning tot he field that needs it

PR #624, opened 2024-04-11
Added fail on warning lint for Javadoc, and rawtypes fixes for Bytes

peter-lawrey:


src/main/java/net/openhft/chronicle/bytes/internal/NoBytesStore.java
```diff
@@ -103,79 +101,79 @@ public boolean tryReserve(ReferenceOwner owner) {
 
     @NotNull
     @Override
-    public RandomDataOutput writeByte(@NonNegative long offset, byte i8) {
+    public NoBytesStore writeByte(@NonNegative long offset, byte i8) {
```
yevgenp, line 104: We're replacing interface with a specific class here and below. That's a regression. Is it a breaking change as well?

yevgenp (approved):

src/main/java/net/openhft/chronicle/bytes/internal/NoBytesStore.java
peter-lawrey: It would be a breaking change if anyone used this implementation directly without recompiling their code, so we wouldn't do it between rolls. However, the implementation rejects the write because it is a placeholder for an empty BytesStore. As such, it's only used internally and should be in an internal package.

PR #627, opened 2024-05-02
X.26

peter-lawrey:


pom.xml
```diff
@@ -18,48 +18,42 @@
...
         <zero.cost.assertions>disabled</zero.cost.assertions>
-        <compiler.argument>-Xlint:deprecation</compiler.argument>
         <sonar.organization>openhft</sonar.organization>
         <sonar.host.url>https://sonarcloud.io</sonar.host.url>
-        <!-- The parent POM defines this plugin and activates it in a profile. Unfortunately this plugin will generate
-        code at @NotNull call-sites that throws IllegalArgumentException rather than NullPointerException. This will
-        break various core tests. Chronicle have raised this issue before with the maintainers of the plugin. See:
-        https://github.com/osundblad/intellij-annotations-instrumenter-maven-plugin/issues/53
-        -->
-        <se.eris.notnull.instrument>false</se.eris.notnull.instrument>
```
JerryShea, line 47: safe to remove?

pom.xml
```diff
@@ -405,6 +367,15 @@
     </scm>
 
     <profiles>
+        <profile>
+            <id>java9-release8</id>
+            <activation>
+                <jdk>[9,)</jdk>
+            </activation>
+            <properties>
+                <maven.compiler.release>8</maven.compiler.release>
```
JerryShea: won't this be defined in the parent pom?

src/main/java/net/openhft/chronicle/bytes/internal/HeapBytesStore.java
```diff
@@ -538,15 +540,24 @@ public void write(
...
+
+                copyMemory0(offsetInRDO, offset, length, src);
             }
         } catch (NullPointerException ifReleased) {
             throwExceptionIfReleased();
             throw ifReleased;
         }
     }
 
+    private void copyMemory0(long offsetInRDO, int offset, int length, byte[] src) {
```
JerryShea, line 553: is this behaviour change?

pom.xml
peter-lawrey: This disables a plugin that has now been removed from the parent.

pom.xml
peter-lawrey: It may have also obscured deprecation issues in Java 17+

src/main/java/net/openhft/chronicle/bytes/internal/HeapBytesStore.java
peter-lawrey: This is an inlining of a removed method in UnsafeMemory.

tgd (approved):

JerryShea (approved):

PR #641, opened 2024-10-15
Remove SDO.write(RDI) methods

peter-lawrey:


src/main/java/net/openhft/chronicle/bytes/HexDumpBytes.java
```diff
@@ -1902,7 +1871,7 @@ public void writeBigInteger(@NotNull BigInteger bi) throws IllegalArgumentExcept
     }
 
     @Override
-    public void writeWithLength(@NotNull RandomDataInput bytes) throws IllegalStateException, BufferOverflowException {
+    public void writeWithLength(@NotNull BytesStore<?, ?> bytes) throws IllegalStateException, BufferOverflowException {
```
tgd, line 1874: @peter-lawrey can we make this change without deprecating first?

tgd (approved):

PR #654, opened 2025-03-21
DEV-132. Delete warmup directories from /tmp

batespj:
Delete warmup directories from /tmp


JerryShea:
Hi @batespj can I make a suggestion for one more change? See [this commit](https://github.com/OpenHFT/Chronicle-Bytes/commit/e3d306bd5aefc31bfe936b35c6edfe2b346bb045) which ensures that the PERF exception handler gets stubbed out during warmup, otherwise you tend to see lots of the below when `warmup` is called, and it pollutes the logs
```
Jvm.perf().on(getClass(), "Took ... ms to add mapping for ....");
```

tgd (approved):

batespj:
Thanks for the suggestion. I agree that stubbing out the PERF exception handler during warmup would help reduce log noise. I think that’s worth addressing, though I’d prefer to handle it in a follow-up rather than hold up this PR. Hope that makes sense.

PR #688, opened 2025-06-04
Harden deserialisation & file mapping; add configurable safeguards

peter-lawrey:
This PR strengthens the library when handling **untrusted input** and interacting with the filesystem. It introduces strict, configurable bounds for array/string reads, canonicalises file paths for mappings, and replaces a few silent failures with safer behaviour and observable logs. It also documents the safer patterns for consumers.

There are **no public API removals**; one small utility addition and a new system property are included.

---

### 1) Safer deserialisation (DoS/OOM resistance)

* **Cap array lengths on read** (all marshalling paths):

  * Added `BytesUtil.checkArrayLength(int len, long remaining)` and `BytesUtil.maxArrayLength()`.
  * Applied checks across `BytesMarshaller` for:

    * `Object[]`, `Collection`, `Map`
    * Primitive arrays: `byte[]`, `int[]`, `long[]`, `float[]`, `double[]`
  * We validate that requested element counts are ≤ configured max **and** ≤ remaining bytes (with per-type element size), otherwise fail fast with `IORuntimeException`.
* **Bounded UTF-8 reads**:

  * `RandomDataInput.readUtf8(...)` and `readUtf8Limited(...)` now validate `offset`, `maxUtf8Len`, and the remaining bytes before decoding. Oversized or out-of-bounds reads raise `BufferUnderflowException` (aligns with **CB-NF-S-001**).
* **Docs**: `api-guide.adoc` now recommends `readUtf8Limited()` when consuming untrusted data and calls out the exception behaviour.

### 2) Configurable input size guardrail

* **New system property:** `bytes.max-array-len` (default **16 MiB**).

  * Central limit for array deserialisation to mitigate accidental OOM or deliberate amplification attacks.
  * Documented in `systemProperties.adoc`.
  * Accessors: `BytesUtil.maxArrayLength()` and guard helper `BytesUtil.checkArrayLength(...)`.

### 3) Path canonicalisation for mapped files

* **MappedFile factory (String path)**: resolves to **canonical path** via `CanonicalPathUtil.of(...)` before mapping.

  * Helps guard against simple path traversal tricks and ensures a single canonical lock key.

### 4) Hardening reflective field access (JDK resilience)

* **ByteBuffer / NativeBytesStore field lookups** now:

  * Catch errors during reflective discovery, **log a warning**, and
  * Fail **explicitly** with `UnsupportedOperationException` if a caller later tries to set addresses/capacities without reflective access.
    *(Prevents partial/undefined states on restricted JDKs or security-managed environments.)*

### 5) Observability & defensive logging

* Replace swallowed exceptions with **debug logs** in a few internal paths:

  * `BytesInternal.appendContent(...)` (vectorised toString checks).
  * `CommonMappedBytes.release(...)` and `AbstractReference.performClose()` now log when releasing a closed resource.
  * `EmbeddedBytes.writePosition()` logs when the backing store is closed.

### 6) Documentation & tests

* Docs updated as above; new property recorded.
* Unit tests adjusted to account for bounds checks (e.g., provide `readRemaining()` in mocks) and string growth logic in `NativeBytesStoreTest`.

---

## Why (threat model & rationale)

* **Unbounded reads** (arrays/strings) can be exploited to trigger **OutOfMemoryError** or large CPU/memory consumption (**DoS**). The new guardrail caps message growth and ties it to actual remaining bytes.
* **Path traversal** via non-canonical paths can defeat naive directory policies and complicate lock coordination. Canonicalising paths is a common, low-cost mitigation step.
* **Silent failures** and **unchecked reflection** can hide incidents or leave objects half-initialised on newer JDKs; explicit failure & logging improves diagnosability and safety.

These changes directly reinforce **CB-NF-S-001** (robust bounds checking) and align with the non-functional security requirements.

---

## Backward compatibility & behaviour notes

* **Array/string reads**: oversized or out-of-bounds inputs that previously slipped through may now fail fast (`IORuntimeException` / `BufferUnderflowException`). This is intentional for security. Legitimate large inputs can be supported by raising `-Dbytes.max-array-len`.
* **MappedFile(String)**: normal cases unaffected; paths are canonicalised before use.
* **Reflection hardening**: on JDKs where direct buffer fields are inaccessible, callers will now get `UnsupportedOperationException` instead of continuing unsafely.
* No API removals; **one new public utility** (`BytesUtil.maxArrayLength()`) added.


peter-lawrey:
Running a Build All Bumped to see is there are any knock on effects

peter-lawrey-admin (approved):

